PR #269, opened 2010-09-02
Small fix in TTURLCache

betamike:
By using UIImage to load image data from a file, the proper scale is set for high resolution screens.  Without this, the high resolution images are displayed at the double the intended size. (I've encountered this issue in conjunction with TTLauncherButton).

### DIFF
--- a/src/Three20Network/Sources/TTURLCache.m
+++ b/src/Three20Network/Sources/TTURLCache.m
@@ -252,8 +252,7 @@ static NSMutableDictionary* gNamedCaches = nil;
 ///////////////////////////////////////////////////////////////////////////////////////////////////
 - (UIImage*)loadImageFromDocuments:(NSString*)URL {
   NSString* path = TTPathForDocumentsResource([URL substringFromIndex:12]);
-  NSData* data = [NSData dataWithContentsOfFile:path];
-  return [UIImage imageWithData:data];
+  return [UIImage imageWithContentsOfFile:path];
 }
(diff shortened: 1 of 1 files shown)


jwang:
Can you fill me in on this? Hi-res meaning @2x for Retina Display? It's a very minor change so if that's the case, I'd like to get it in but want to make sure I have a test case to verify it.

betamike:
Yes hi-res meaning properly loading @2x retina display images.  After doing some additional testing, I found that this will not automatically load the @2x file given the lower resolution file path.  This will properly set the image scale if you pass the @2x file path though.

jwang:
so we still need something that can automatically load the @2x given a path to the low res (assuming the high res is available in the same path, per how iOS auto detection works.)

betamike:
Yep.  Unfortunately, the automatic @2x detection (or at least setting the proper scale) is broken with the UIImage methods that allow you to load an image from a filepath (this was fixed in 4.1).  In order to get this working properly, it will probably require a workaround.

http://stackoverflow.com/questions/3739812/how-to-get-uiimage-imagewithcontentsoffile-and-high-res-images-working

jwang:
interesting. I've run into some other bugs that were in 4.1 but fixed in 4.2 as well. Thanks. Will take a look at the SO question.

jwang:
added as 41b2a029475e8b0e407feb24f4be5ea318ef4220

PR #273, opened 2010-09-08
Fixed TTRequestLoader for local urls

airportyh:
TTRequestLoader (and therefore photoviewer) didn't work for local urls. This commit fixes that.


jwang:

PR #275, opened 2010-09-11
Misc changes to Three20

bensummers:
I needed to make some tweaks to Three20 for my app, and just wondered if any of them might be useful for the core.

The ones you might want are the start of pasteboard support for copying table cells, and a fix for margin and padding calculations in TTStyledTextTableItemCell so it displays correctly inside a cell with rounded corners, and optional display of an image in TTTableStyledTextItem.

The other two are probably just for my app, and make things a little more consistent in display. They would probably break other apps, but I thought you might want to see them anyway.

Thanks!


jwang:

jwang:
Hi @bensummers I pulled in a few changes that removed some things. I like the pasteboard and want to pull that in. I don't use the pasteboard too much, does it not support styled text? Only plaintext works? Thanks! :)

bensummers:
My pasteboard implementation only copies plain text -- I was only writing what I needed for my app.

PR #277, opened 2010-09-12
NSArrayAdditions perform:selector is unsafe

mgc:
perform:selector: in NSArrayAdditions contains a subtle crash bug.

See http://github.com/facebook/three20/issues/issue/276


mgc:
[Core] Retain objects in NSArray perform: methods.

Performed copy before enumerating, but didn't account for
non-retaining delegate arrays. Result was a subtle crash.

Closed by a1555fb786180290333accde2ec850e868b58704
Closed by a1555fb786180290333accde2ec850e868b58704

PR #285, opened 2010-09-27
The query dictionary gets persisted/restored just like the URL

fpotter:
I've found it really helpful to have the query dictionary be persisted in the same way as the URL, so that when controllers get restored, all the original parameters are available.

The nice thing about having this handled inside of three20 (vs. having each controller do it in their own restore/persist) is that the original parameters are available in the init method.  That is, you don't have to wait until after restoreView to figure out what the original params were...


fpotter:
Yeah, I clearly submitted this pull request wrong.  Will resubmit later with just the one relevant commit.

PR #312, opened 2010-11-02
TTStyledLayout and large images

jarnoan:
If an image is wider than the width of the layout, the layout width becomes bigger, which is incorrect in my opinion. This change causes the image to be scaled down to fit the width of the layout. 

This was a quick fix and only works when there is no padding. This was enough for my case, but in general case this would need a bit more work, but I was afraid of touching anything more than I needed to.


jwang:

jamesgpearce:

PR #313, opened 2010-11-02
TTStyledTextTableItemCell layout fix for margin

jarnoan:
TTStyledTextTableItemCell didn't use margin for width calculations correctly, and this should fix that, I think.


vguerci:
had the same issue... and wrote exactly the same fix! So I confirm it works...
I wished I looked at pull requests first, that would have saved me some time...
...or better that pull request has been accepted :)

diederich:
+1 for merging

vguerci:
for information, this fix is a must have if you want to use a TTStyledTextTableItemCell in a table with grouped style, otherwise text is not aligned and cell is over the rounded corners on top and bottom of sections

jwang:

jeanregisser:
It seems I have the same fix here: https://github.com/facebook/three20/pull/484

PR #325, opened 2010-11-07
V1.0

steipete:
hi jeff,

so here's the commit with the right icons now.
i've no idea how to get rid of my last commit, but you can just cherry-pick it or dl' the two @2x files.

thanks for pointing out the blurryness... would have overlooked it. (and yet I love the details!)

best, 
- Peter


steipete:
Add 2x variants of forwardIcon/backIcon.

Replaced original icons with copies from UIKit.

Closed by 390f5b6b54bb25f4d614876a9b4e5870a4839460

PR #339, opened 2010-11-17
Fix for #338 - Make TTStyledTextLabel use the correct font when rendering bold and italic text.

fpotter:
This is to make sure <i>some text</i> uses the italic version of whatever font the TTStyledTextLabel is supposed to use, and likewise for bold text.  Previously, it would just always use the [italic|bold]SystemFont.

Included in here is a table of iOS fonts and the specific font names for the the bold, italic, and bold & italic versions.  Not all fonts are listed - just the ones with multiple versions.

The original issue:
https://github.com/facebook/three20/issues/issue/338

### DIFF
--- a/samples/TTCatalog/Classes/StyledTextTestController.m
+++ b/samples/TTCatalog/Classes/StyledTextTestController.m
@@ -115,6 +115,18 @@ Both line break characters\n\nand HTML line breaks<br/>are respected.";
   //label1.backgroundColor = [UIColor grayColor];
   [label1 sizeToFit];
   [self.view addSubview:label1];
+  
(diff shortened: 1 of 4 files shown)


jwang:
I don't know about including the table and using for a look-up. I understand the premise, but the fonts can and do change. For example, iPad 4.2 supports 57 different fonts and iPhone 4.2 supports 40. I don't know why that is, but such is the case. I'll write up a dynamic lookup solution for this.

fpotter:
Awesome - something dynamic would be great.  Just wasn't sure how it could be done w/out matching against the names.

jwang:
closing this in favor of dynamic font lookup.

PR #340, opened 2010-11-18
TTStyledTextLabel - Some text isn't being aligned on the same baseline as other text.

fpotter:
See this screenshot to understand the issue (look for the funky alignments):
http://fpotter_public.s3.amazonaws.com/bad_font_size_16.png

You'll notice I'm not using Helvetica, which is I think the reason I came across this issue in the first place.  Helvetica just seems to work.  And, I'm also using bold and italic text in non-Helvetica fonts, which is enabled by this change: https://github.com/facebook/three20/pull/339

The issue is is in `TTStyledLayout.m`, in `layoutText:container:`.  There are five places in there where `addFrameForText` gets called, and in most cases the height for the frame is taken from the `_lineHeight` variable.  If we change the last case to use `_lineHeight` as well, things start working.  I wonder if this was just an accidental omission?

To go deeper...

You would think it would just work as-is.  `_lineHeight` should be equal to `[font ttLineHeight]` but they're not for at least some fonts and point sizes.  For example, in the case I was debugging, `_lineHeight` was 20.0 but `[font ttLineHeight]` was 19.57..something.  And, when ttLineHeight is less than `_lineHeight`, TTStyledLayout tries to compensate by adding in offsets in `breakLine`, but it adds the wrong offsets.

See....

```
if (_lineFirstFrame.nextFrame) {
  TTStyledFrame* frame = _lineFirstFrame;
  while (frame) {
    // Align to the text baseline
    // XXXjoe Support top, bottom, and center alignment also
    if (frame.height < _lineHeight) {
      UIFont* font = frame.font ? frame.font : _font;
      [self offsetFrame:frame by:(_lineHeight - (frame.height - font.descender))];         
    }
    frame = frame.nextFrame;
  }
}
```

I'm still a little confused, but my bug is fixed :-)  Feedback welcome!

### DIFF
--- a/src/Three20Style/Sources/TTStyledLayout.m
+++ b/src/Three20Style/Sources/TTStyledLayout.m
@@ -740,7 +740,7 @@
                                   lineBreakMode:UILineBreakModeWordWrap];
 
         [self addFrameForText:lines element:element node:textNode width:linesSize.width
-             height:linesSize.height];
+             height:_lineHeight ? _lineHeight : [_font ttLineHeight]];
         _height += linesSize.height;
         break;
       }
@@ -757,7 +757,7 @@
         frameWidth = [[text substringWithRange:NSMakeRange(frameStart, stringIndex - frameStart)]
                       sizeWithFont:_font].width;
         [self addFrameForText:line element:element node:textNode width:frameWidth
-              height:[_font ttLineHeight]];
+              height:_lineHeight ? _lineHeight : [_font ttLineHeight]];
         frameWidth = 0;
       }
     }
--- a/src/Three20Style/Sources/TTStyledLayout.m
+++ b/src/Three20Style/Sources/TTStyledLayout.m
@@ -740,7 +740,7 @@
                                   lineBreakMode:UILineBreakModeWordWrap];
 
         [self addFrameForText:lines element:element node:textNode width:linesSize.width
-             height:_lineHeight ? _lineHeight : [_font ttLineHeight]];
+             height:linesSize.height];
         _height += linesSize.height;
         break;
       }


mmorearty:
I, too, would love to see baseline issues get fixed, but unfortunately I think your proposed fix causes other problems, Fred.  I have a sample app that makes a table whose data source include this:

```
- (id)init {
  if (self = [super init]) {
    self.items = [NSArray arrayWithObjects:
      [TTTableStyledTextItem itemWithText:[TTStyledText textFromXHTML:@"hello <span class='bigger'>world</span> it's me"]],
      nil];
  }
  return self;
}
```

My stylesheet defines a class called "bigger" that is just a bigger font:

```
- (TTStyle*)bigger {
  return [TTTextStyle styleWithFont:[UIFont systemFontOfSize:36] next:nil];
}
```

Here is what is looks like [without](http://www.morearty.com/three20/issues/340/before.png) your change and [with](http://www.morearty.com/three20/issues/340/after.png) your change.  As you can see, it looks better without your change.  (It still doesn't look quite right -- the baseline of the word "world" isn't quite lined up with the other text -- but that is true in both images, and doesn't have anything to do with your code.)

mmorearty:
It turns out that a long time ago, Joe Hewitt added the specific change you are proposing here, and then later he removed it.

In revision 1196ab4fc37795cc2ea2c97799d68227af50dc99 on 4/28/09, he added `_lineHeight ? _lineHeight :` to two places in the code, without much explanation (the comment on that commit seems to have more to do with other changes made in the same commit).  Then in revision 2b0f3923b78e6672a2261f0aaffd5205a4a11c64 on 8/9/09, he backed out one of those changes, with the comment, "Fixed: vertical alignment issues with small font after large font."  Which is exactly the case I was pointing out in my previous comment.

I actually think that `_lineHeight ? _lineHeight :` should not have been added to _either_ location in the code.  I'm going to log a separate bug on that.

But I also found, as you did, that the code in `breakLine` that attempts to align the baselines is incorrect.  I'm going to log that too; I have a fix, and I'm hoping it will fix your case.  Stay tuned...

jwang:
mmorearty, did you log those issues separately? If so, what are those issue numbers. I'm looking through the list and can't find them.

mmorearty:
John, sorry for the delay.  I have now submitted the pull requests: facebook/three20#377 and facebook/three20#378

Fred, I would be interested to know if these patches (especially the first one) fix your issue.

jwang:
Thanks mike. scheduled for 1.0.4 release. I'll be pulling this in for testing this week.

PR #374, opened 2011-01-25
Update build settings to use latest iOS

benr75:
The current master branch of three20 is set to iOS 4.1, in the latest iOS SDK (as I am sure you know) there is a new setting for simply picking the latest iOS SDK for compiling.  These two commits make that change, I would think this is preferred moving forward.

~ Ben


jwang:
These changes have been made already. They'll be changed in the next release shortly. Thanks.

PR #376, opened 2011-01-28
Make TTWebControllerDelegate do a little more

hongrich:
TTWebControllerDelegate currently does nothing. This commit makes it do a little more.


hongrich:
Never mind. This seem like a pretty useless change.

PR #387, opened 2011-02-07
Improved URL encoding for stringByAddingQueryDictionary:

pazustep:
Currently, -stringByAddingQueryDictionary: only performs two simple substitutions before returning the new string.

Since I believe the output of this method will should more than likely be used to construct URIs, further escaping is needed to account for non-ASCII characters and a number of reserved characters that shouldn't be used in URIs.

This commit adds a new method to NSString(TTAdditions) to perform such escaping, and uses the new method from -stringByAddingQueryDictionary:.

### DIFF
--- a/src/Three20Core/Headers/NSStringAdditions.h
+++ b/src/Three20Core/Headers/NSStringAdditions.h
@@ -46,7 +46,16 @@
 - (NSDictionary*)queryContentsUsingEncoding:(NSStringEncoding)encoding;
 
 /**
- * Parses a URL, adds query parameters to its query, and re-encodes it as a new URL.
+ * Returns a string will all non-ASCII characters, reserved characters for URIs (;/?:@&=+$,),
+ * and a few other trouble characters (!'()*\n\"<># \t) replaced with escape sequences using
+ * percent encoding. Strings returned by this method should be safe to be used as any URI
+ * component, including query parameters.
+ */
+- (NSString*)stringByEncodingAsURL;
+
+/**
+ * Parses a URL, adds query parameters to its query, and re-encodes it as a new URL. Both
+ * keys and values in the query dictionary are encoded using -stringByEncodingAsURL.
  */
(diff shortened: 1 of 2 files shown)


tonklon:
This is something I have been missing, too. Thanks a lot for your work.

###Few thoughts on this:

stringByEncodingAsURL should take an NSStringEncoding as argument and use this for the encoding instead of hardcoding NSUFT8StringEncoding.

I think stringByEncoding__For__URLUsingEncoding: or stringByURLencodingStringUsingEncoding: do better describe the method.

stringByAddingQueryDictionary: has been in Three20 for a long time, changing it's behavior is likely to break a lot of code, since people have been used to urlencode their arguments before sending this message, and will therefore end with doubly encoded values. And it does not take an encoding as parameter, as discussed above.

IMHO we would need a new method called stringByAddingUrlencodedQueryDictionary: usingEncoding:

jwang:
Thanks guys. Will take a look and see how we can best get this in.

tonklon:
I could implement it, if you are fine with my suggestions and pazustep doesn't do it.

jwang:
Thanks tonklon. It's not so much a question of implementing. That's easy enough. I just want to make sure that in all of the places we're calling the method, it's calling the proper one and with the correct parameters. Also, when we add it, we need to add the corresponding unit test to make sure we're compliant. :)

tonklon:
This is one reason why I was suggesting to not change any existing methods, but instead only add new ones. And by implementing I mean implement the method, document it and implement unit tests.

I must admit I haven't looked were `stringByAddingQueryDictionary:` is used inside Three20 Code.

jwang:
@tonklon Did you get around to implementing this? I need @pazustep to send it and target development branch. Let me know if you want take it and build your changes on top of his and submit it instead. Thanks! :)

tonklon:
I just created pull request #644 which addresses this.

IMHO this request schooled be closed and not merged.

PR #390, opened 2011-02-11
Support sending application/x-www-form-urlencoded forms

jdewind:
Currently the TTURLRequest always submits a POST using multipart/form content type in order to support data and images out of the box. However, there are times when this is not necessary (most of the time) and some APIs stipulate the "Content Type" for a particular request -- returning an error if the content type is not respected.

This pull request contains a small set of changes that allows a URL request to be configured with a non "multipart" post body. By default, the request is multipart in order to be inline with the current expectations of a POST under TTURLRequest  -- which is every post contains a multipart/form body.

For example,

  TTURLRequest *urlRequest = [[[TTURLRequest alloc] initWithURL:@"http://www.google.com" delegate:nil] autorelease];

  [urlRequest setHttpMethod:@"POST"];

  urlRequest.multiPartForm = NO;

I have tests for these changes. However, I wrote them using a different test framework. The gist can be found [here](https://gist.github.com/823018).

### DIFF
--- a/src/Three20Core/Headers/NSStringAdditions.h
+++ b/src/Three20Core/Headers/NSStringAdditions.h
@@ -51,6 +51,11 @@
 - (NSString*)stringByAddingQueryDictionary:(NSDictionary*)query;
 
 /**
+ * Returns a URL Encoded String
+ */
+- (NSString*)urlEncoded;
+
+/**
  * Returns a string with all HTML tags removed.
  */
 - (NSString*)stringByRemovingHTMLTags;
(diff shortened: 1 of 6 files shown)


jwang:
Thanks. Will take a look. I think there's another issue logged that's related. I'll relate them later today. :)

jdewind:
Great, I only noticed the issue recently when I encountered an unusually strict Web API.

jwang:
can you change your commit range to development branch? It looks like there's conflicts here.

jdewind:
See Pull Request #593

jwang:
awesome thanks!

PR #392, opened 2011-02-12
set rootViewController of window

steipete:
THIS IS JUST AN INSPIRATION

I believe the UIWindow's rootViewController should be set by TTNavigator automatically.
This is a one-line-attempt to fix that.

If three20 wants to stay 3.x compatible you should add respondsToSelector to the call (also, remove my hack comment)


jwang:
Thanks. Will take a look.

PR #393, opened 2011-02-12
Language Change; German Pull To Refresh is too long

steipete:
This changes "Herunterziehen" into "Ziehen" which is basically the same and fits perfectly.


diederich:
applied in 244ef2b9f794b7411452

PR #415, opened 2011-02-19
Cache policy

savanah:
This allows an image loader to force an update using cache policy TTURLRequestCachePolicyNoCache

(Useful for advertising banners, where we want to force a new ad to load.)


jwang:
Thanks!

Note to self: seems related to #19

jwang:
closed bef397b6b0d7a1843473fb5b9a0d765740c95a6e

PR #417, opened 2011-02-19
Removed some static analyzer warning

charlymr:
Hi all, 
some static analyzer warning with LLVM GCC 4.2
also noticed three20 library is still compiling in GCC 4.2 which is on it's way out.
Regards,

Denis


jwang:
Ah. Thanks. This is actually all going in for the new Style Guide. These are actually all in our style-refactor branch that we're updating.

As for the GCC 4.2 vs LLVM GCC 4.2 - Still working through making sure everything is ok. That change is actually related to #365.

PR #435, opened 2011-02-26
Fix for Memory Leak in TTTableView.m

steipete:
Sorry for the pull request mess; I'm talking about this small commit:

https://github.com/steipete/three20/commit/0a849cb0e6faa0aabb7d84708484dfde232f60b2

Found them with Instruments and fixed the leaks.


diederich:
Hm, not sure on how to comment on this pull request. There seem to be some more commits involved. Mind fixing this?

Besides that, did you check why _originShadow is handled differently? While it seems to be a weak-ref only, shouldn't this be handled the same way?

steipete:
_doh_ github somehow merged this with earlier pull requests...
i'll fix that, sorry.

PR #444, opened 2011-03-03
Xcode 4 support for ttmodule.py, class attribute support for button node in TTStyledTextParser, userInfo property support for TTLauncherItem

taknology:
1. Updated TTStyledTextParser.m to allow for class attribute to be assigned to button node types
2. Updated TTLauncherItem.h,TTLauncherItem.m to add userInfo property for assigning user defined items for retrieval when buttons are edited, touched, moved, etc.
3. Updated Pbxproj.py to work with Xcode 4. Retained backward compatibility with Xcode 3.

### DIFF
--- a/src/Three20Style/Sources/TTStyledTextParser.m
+++ b/src/Three20Style/Sources/TTStyledTextParser.m
@@ -199,6 +199,7 @@
 
   } else if ([tag isEqualToString:@"button"]) {
     TTStyledButtonNode* node = [[[TTStyledButtonNode alloc] init] autorelease];
+    node.className =  [attributeDict objectForKey:@"class"];
     node.URL =  [attributeDict objectForKey:@"href"];
     [self pushNode:node];
 
--- a/src/Three20UI/Headers/TTLauncherItem.h
+++ b/src/Three20UI/Headers/TTLauncherItem.h
@@ -30,8 +30,10 @@
   NSString*       _badgeValue;
 
   BOOL            _canDelete;
+    id			  _userInfo;
 }
 
+@property (nonatomic, assign) id userInfo;
 @property (nonatomic, assign) TTLauncherView* launcher;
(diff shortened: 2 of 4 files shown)


jwang:
argh. can you make each one a separate pull request?

taknology:
Wow, really?  Ok.  Let me see what I can do. Each enhancement right?  So TTStyledTextParser.m in one, TTLauncherItem.h and TTLauncherItem.m in another and Pbxproj.py in a third?

jwang:
If you can, it would make things easier to deal with. Otherwise, I can deal with it on my end.

taknology:
I'm really new to Git commits so this may take a bit.  I'll give it a whirl.

jwang:
no worries. we're not git experts either. :)  For XCode4, are you using Workspaces? I'm thinking that may be a better route for us to go in the future. I like this in the interim, but it seems XCode4's workspaces are really powerful. :)

taknology:
Not currently.  This change was made with the assumption that Three20 would be added to a fresh now project. Have not looked at support for workspaces yet. I come from the workd or Windows development so if Xcode4's workspaces work similar to the way Visual Studio handles workspaces that it's going to be sweet.

taknology:
Ok.  I have submitted a new pull request. This could get tedious as I can only submit one pull request at a time.  Once this pull request is approved I can submit the other two.

jwang:
Thanks! You should be able to submit as many as you want. There shouldn't be anything stopping you from submitting more pull requests for other changes.

taknology:
Ok, cool.  I figured it out.  Total newbie with Git, but I must say I like the concept.  I don't know how many times I have seen SVN repositories get corrupted due to poorly managed commits.

jwang:
It's a gateway drug. Beware, you'll likely go nuts if you have to deal with SVN ever again. :)

PR #457, opened 2011-03-06
Fix issue #251: TTBaseNavigator not respecting modalPresentationStyle and modalPresentationStyle (iPad)

kemenaran:
This patch implements the fix suggested by niblh in the issue #251.

### DIFF
--- a/src/Three20UINavigator/Sources/TTBaseNavigator.m
+++ b/src/Three20UINavigator/Sources/TTBaseNavigator.m
@@ -304,6 +304,9 @@ UIKIT_EXTERN NSString *const UIApplicationWillEnterForegroundNotification __attr
 
   } else {
     UINavigationController* navController = [[[[self navigationControllerClass] alloc] init] autorelease];
+    navController.modalTransitionStyle = transition;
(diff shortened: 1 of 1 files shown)


jwang:
thanks, but we still need that test case. :)

kemenaran:
Here is one example:

```
@interface FooController : UIViewController
@end

@implementation FooController
- (id)initWithNibName:(NSString *)nibNameOrNil bundle:(NSBundle *)nibBundleOrNil {
    self = [super initWithNibName:nibNameOrNil bundle:nibBundleOrNil];
    if (self) {
        self.modalPresentationStyle = UIModalPresentationFormSheet;
    }
    return self;
}
@end

@interface AppDelegate : NSObject <UIApplicationDelegate>
@end

@implementation AppDelegate
- (void)applicationDidFinishLaunching:(UIApplication *)application {
    [[TTNavigator navigator].URLMap from:@"tt://modal" toModalViewController:[FooController class] transition:UIModalTransitionStyleFlipHorizontal];
    TTOpenURL(@"tt://modal");
}
@end
```

Do you need a full unit-test implementation?

jwang:
closed by #591

PR #460, opened 2011-03-08
Fix gitignore

rogchap:
Added Python generated files (*.pyc) to .gitignore strings


diederich:
+1 for merging, seen them here, too.
Should we explicitly add them or is the *.pyc rule safe?

jwang:
lol. i should commit it then. i didn't commit my ignore file even though i added it. i'll push it now.

jwang:
[REPO] update gitignore to add .pyc files. closed by 1bf9e345292f170d1d583c3a766ea75435174afc

PR #461, opened 2011-03-08
Fix backgroundColor in tableview cell labels

rogchap:
When setting a cell backgroundView in UITableViewDelegate the Labels within a cell (i.e. TTTableSubtitleItemCell) takes on the cell backgroundView as the UILabel's background view.
Setting the backgroundColor of these labels fixes this.
I've added a backgroundTextColor to the TTDefaultStyleSheet for this. The aim would be to utilize this for other UILabels. Most applications require the backgroundColor to be clearColor apposed to the default whiteColor


rogchap:
This fixes issue 73: https://github.com/facebook/three20/issues/#issue/73

jwang:
thanks! you can link to issues just by using the # with the number after, like: #73

PR #464, opened 2011-03-10
Fix for TTTextEditor doubling first newline

vguerci:
I can't believe that this fix is so simple and that bug hanging around for years...
I suspect this `\n` addition to be there for one strange reason.
Then I would recommend more testing, but that does fix the TTCatalog table control item sample.
Related bugtracker issue : #317

### DIFF
--- a/src/Three20UI/Sources/TTTextEditor.m
+++ b/src/Three20UI/Sources/TTTextEditor.m
@@ -237,7 +237,7 @@ static const CGFloat kUITextViewVerticalPadding = 6;
     [self createTextView];
     _textField.hidden = YES;
     _textView.hidden = NO;
-    _textView.text = [_textField.text stringByAppendingString:@"\n"];
+    _textView.text = _textField.text;
     _internal.ignoreBeginAndEnd = YES;
     [_textView becomeFirstResponder];
     [self performSelector:@selector(stopIgnoringBeginAndEnd) withObject:nil afterDelay:0];


jkubicek:
I can't test this fix right now, but I'm pretty confident that the issue is much more subtle than just removing the additional `\n`. Have you tested this change to make sure that the view reliably swaps between the UITextView and UITextField subclasses?

vguerci:
I tested this on TTCatalog sample in table control items view.
This code is only called once when UITextfield swapped by TTTextview.

jwang:
this does need more testing. :)

dimitar-lazarov:
I don't agree with the vguerci's fix. It doesn't work when the first typed char is new line. It looks like the problem here is that textFieldShouldReturn method is called before the new line char is passed to first responder and UITextView receives it when it is reassigned in the same method - so the new line becomes doubled. 

I think better solution is to return NO in textFieldShouldReturn method, so the new line char will not be passed to first responder. Also I think that will be better if UITextView and UITextField are swapped not when Return is pressed but after first char is typed in UITextField, and swapped back when user delete all chars. This will fix also the issue when user changes cursor position in the middle of typed text and press Return.

dineshRajpoot:
I agree with intruder. Returning no in textFieldShouldReturnMethod helped me in solving this issue.

# 

diff --git a/src/Three20UI/Sources/TTTextEditorInternal.m b/src/Three20UI/Sources/TTTextEditorInternal.m
index 2db4ce9..afbf6b8 100644
--- a/src/Three20UI/Sources/TTTextEditorInternal.m
+++ b/src/Three20UI/Sources/TTTextEditorInternal.m
@@ -216,6 +216,7 @@ TT_FIX_CATEGORY_BUG(TTTextEditorInternal)
   }

>  [_textEditor performSelector:@selector(didChangeText:) withObject:(id)YES];
>  return NO;  //Added this line
> 
>  if ([_delegate respondsToSelector:@selector(textEditorDidChange:)]) {
>    [_delegate textEditorDidChange:_textEditor];

jamesgpearce:

PR #477, opened 2011-03-14
Tturlrequest mod

crino:
Working on my FriendFeed client (FFHound) i modified the network library, in spec:
1. now if the error is generated by an error code in 'connectionDidFinishLoading:(NSURLConnection *)connection', i add the response data into the NSError in 'kTTErrorResponseDataKey' so i can retrive the error' server later. This because the server reply with an error code (401, 403 etc..) but the response contains details on error;
2. using the 'official' post body in TTURLRequest all requests failed remaining 'hanged' so i 'rewrote' the method removing the "\r\n" in beginLine variable and adding the endLine variable.

Thanks


jwang:
simple enough. thanks.

crino:
hi jwang, sorry for late reply. anyway, thanks to integrate my 'pull request' ;)

PR #480, opened 2011-03-16
Show menu animation

jeffarena:
The default showMenu/hideMenu animation causes the table cell to slide out to the left and back in from the left, which is awkward in response to a standard swipe from left to right.  This change causes the cell to slide out to the right and back in from the right, which in my opinion is a bit more intuitive.


jwang:
thanks. will test it with ttcatalog

PR #482, opened 2011-03-17
[UI] Fix TTTextStyle shadow flipping bug

jeanregisser:
Hi,

This was originally submitted as part of the fixes in https://github.com/facebook/three20/pull/307
To make it easier for you to merge it, it is now based off the facebook:development branch.

Thanks.

### DIFF
--- a/src/Three20Style/Sources/TTTextStyle.m
+++ b/src/Three20Style/Sources/TTTextStyle.m
@@ -22,6 +22,7 @@
 #import "Three20Style/UIFontAdditions.h"
 
 // Core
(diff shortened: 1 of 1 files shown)


diederich:
What about factoring out that check to one of the TTCore headers and use in in the various places?

jeanregisser:
Yes that sounds like a good idea.

jeanregisser:
Why not merging this in?

jwang:
That's a good question. Why did I close this. Let's re-open it and put it into next week's push. I must've thought I merged it in already.

PR #483, opened 2011-03-17
[UI] Fix TTTableImageItemCell default image bug

jeanregisser:
Hi,

Here is a really small fix ;)
It fixes TTTableImageItemCell not displaying the default image when imageURL is nil.

This was originally submitted as part of the fixes in https://github.com/facebook/three20/pull/307 

To make it easier for you to merge it, it is now based off the facebook:development branch.

Thanks

### DIFF
--- a/src/Three20UI/Sources/TTTableImageItemCell.m
+++ b/src/Three20UI/Sources/TTTableImageItemCell.m
@@ -147,7 +147,7 @@ static const CGFloat kDefaultImageSize = 50;
     ? image.size.height
     : (item.imageURL ? kDefaultImageSize : 0);
 
-    if (_imageView2.urlPath) {
+    if (_imageView2.urlPath || image) {
       CGFloat innerWidth = self.contentView.width - (kTableCellHPadding*2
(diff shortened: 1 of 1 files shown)


diederich:
Thanks!

'image' is set depending on the exactly this condition a few lines earlier, so what about checking for nil != image instead?

jeanregisser:
If we just check for nil != image, then the case where there's an imageURL set but no default image and no cached image for imageURL, will be wrong:

-> _imageView2 won't be displayed.

diederich:
yep, sorry, that makes sense!

changing _imageView2.urlPath to item.urlPath might make it somehow easier to read, given that the other checks depend on item.urlPath, too (and they should be the same anyway)

+1 for merging

rogchap:
+1 for merge
Will this effect other table cells like TTTableSubtitleItemCell?

jeanregisser:
I haven't checked if TTTableSubtitleItemCell or other cells displaying images are affected by the problem I fixed in TTTableImageItemCell.

PR #487, opened 2011-03-17
Set the originalNavigatorURL when using hash-URLs (tt://controller#5) 

kemenaran:
When a hash-URL (like `tt://controller#(method:)`) is opened, the TTNavigator forgets to set the originalNavigatorURL of the controller when creating it.

That means if another hash-URL pointing to the same controller is called after that, the TTNavigator won't recognize the current controller, and will create and push a new one again.
#### Steps to reproduce

```
/* 1. */ [map from:@"tt://controller#(setSelectedPage:)" toViewController:[TTPagesController class]];
/* 2. */ TTOpenURL(@"tt://controller#1");
// later…
/* 3. */ TTOpenURL(@"tt://controller#5");
```
#### Expected behavior

On `3.`, the existing controller instance receives a `setSelectedPage:` message.
#### Observed behavior

On `3.`, a new controller is instanciated. This is because as the originalNavigatorURL wasn't set on the original instance, the TTNavigator can't figure out that `tt://controller#5` actually maps to a controller previously instanciated with `tt://controller#1`.

### DIFF
--- a/src/Three20UINavigator/Sources/TTBaseNavigator.m
+++ b/src/Three20UINavigator/Sources/TTBaseNavigator.m
@@ -692,13 +692,16 @@ __attribute__((weak_import));
     } else {
       id object = [_URLMap objectForURL:baseURL query:nil pattern:pattern];
       if (object) {
+        UIViewController *controller = nil;
         id result = [_URLMap dispatchURL:URL toTarget:object query:query];
         if ([result isKindOfClass:[UIViewController class]]) {
-          return result;
+          controller = result;
 
         } else {
-          return object;
+          controller = object;
         }
+        controller.originalNavigatorURL = baseURL;
+        return controller;
 
       } else {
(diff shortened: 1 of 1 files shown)


teehemkay:
Isn't this the  normal behavior?

To re-use the controller shouldn't  1. above instead be (please note that use "toSharedViewController" in lieu of "toViewController"):

 [map from:@"tt://controller#(setSelectedPage:)" toSharedViewController:[TTPagesController class]]; 

?

OTOH #fragment URL are indeed not persisted by the TTNavigator because  originalNavigatorURL is not set in the case of a #fragment URL. Which may or may not qualify as a bug.

But to persist the #fragment URL, originalNavigator should be set to the full #fragment URL not the base URL.

jwang:
That's correct @exidia. It is the normal behavior.

PR #488, opened 2011-03-19
Fixed wrong spelling

meap:


jwang:
Ouch. That's a pretty big issue. Thanks.

PR #489, opened 2011-03-20
iOS Deployment Target of 3.1 and LLVM as the default compiler. Related: suppressing some LLVM compiler warnings

jverdi:
Justification below for moving to a 3.1 Deployment Target and LLVM Compiler.
Also, suppressed warnings about "assignments as a condition without parenthesis" when compiling with LLVM 2.0


jverdi:
Suggesting that three20 raise the iOS Deployment Target to 3.1, and start using LLVM as the default compiler.

The reason for this is that currently, three20 does not run on iOS 3.x devices. Symbolic linking of the following objects is impossible at runtime:  
- UIPopoverController  
- UIPopoverControllerDelegate  
- UISplitViewController  

As I understand it, based on my reading of Marco Arment's post on the matter (found here: http://www.marco.org/1648550153) there are several methods to work around this issue and maintain 3.0 as the iOS Deployment Target:  
1) Use NSClassFromString() to avoid these symbolic references. Unfortunately, that method won't work for three20 because it creates a subclass of UISplitViewController. Marco also discusses the possibility of building the subclass at runtime, but that seems like far more trouble than it's worth.  
2) Weak link the entire UIKit Framework. The performance implications that result would likely make this solution prohibitive.

If three20 is willing to go to a 3.1 Deployment Target, there are much more elegant solutions. Since iOS 4.2, Apple has added the ability to weak link only specific classes rather than entire frameworks.  
- The method for doing this is explained here: http://www.sealiesoftware.com/blog/archive/2010/4/8/Do-it-yourself_Objective-C_weak_import.html

Fortunately, if Base SDK is greater than 4.2, and Deployment Target is 3.1, either of the two LLVM compilers will do all of this weak linking automatically. Therefore, the reasonable solution to me would be to target 3.1 minimum and switch to LLVM as a default.

Justification for going to 3.1:  
All of the major three20 users listed on three20.info target 3.1 or greater already. This includes SCVNGR (3.1), Pulse (3.2), Meetup (3.1.2), and Posterous (4.0). One notable exception is facebook, though I imagine they are not using the latest three20 as a base as they have no need for these iPad specific classes.

One closing thought: I'm curious whether it would be feasible to move all iPad related code to a separate three20 extension. While the iPad specific code is not the only thing with that may require weak-linking in the future, it would be a logical separation and puts this issue off for at least for the time being.

jwang:
I've been looking at this for quite a while now. And I've read Marco's comment as well. I've tried his option and it hasn't quite worked as nicely for me. Even with the options he had given, I was not able to run TTCatalog on a 3.1 device without weak-linking UIKit. I don't think we'll be giving up on 3.0 just yet either. I would like to utilize LLVM a bit more and the LLVM GCC 4.2 is an option as well. I have pretty much these same changes on a local branch and it's an issue that's been up to discussion before.

jwang:
Note. Related to #365

jverdi:
If giving up 3.0 isn't in the cards, what are your thoughts on separating the iPad specific code to a separate optional extension?

I think that may allow most to get on with publishing iPhone/iPod apps with 3.x targets for the time being.

jwang:
Jeff has some top secret iPad branch on his computer that we haven't seen yet. I'll follow up with him on what his iPad branch has. It wouldn't be a bad thing to have it as an extension, since Three20 is trying to modularize everything slowly. So that part will likely pan out in the future.

jverdi:
Just want to note here: after running into some problems of my own, and re-reading Marco's post, I noticed that LLVM GCC must be used with iOS targets < 4.0. LLVM Compiler 2.0 won't work on older 3.x versions.

jwang:
That's correct. Sorry, I thought that was mentioned, but guess not. I'm going to close this for now as we're not quite ready to give up on pre- iOS 4.0 yet.

PR #490, opened 2011-03-23
Development

levous:
dateAtMidnight was presenting as a major performance hog in our application.  I found that using NSDateComponents accomplishes the same thing in about 1/7 the cpu time.


jwang:
nice. thanks!

PR #494, opened 2011-03-23
Core Data

MattNewberry:
Spoke with jwang392 on IRC about this, it's not perfect, but a baseline I'm using for the Shopify mobile app. 

It works quite well actually until trying to invalidate the model and refresh the table. 

I had to wrap the TTTableViewController delegate methods for update, insert, and delete in @try blocks to prevent crashes, and manually try and go from there... Not good! 

The table typically will load the visible rows without issue in this manner, however when you try to scroll down, it's all blank data. 

I use this heavily with my ActiveRecord port to Objective-C that I wrote, https://github.com/MattNewberry/ObjectiveRecord

Makes it much easier to work with a remote Rails datasource.

### DIFF
--- a/src/Three20UI/Headers/NSFetchedResultsDataSource.h
+++ b/src/Three20UI/Headers/NSFetchedResultsDataSource.h
@@ -1,0 +1,65 @@
+//
+//  NSFetchedResultsDataSource.h
+//  Shopify_Mobile
+//
+//  Created by Matt Newberry on 11/15/10.
+//  Copyright 2010 Shopify. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+
+@interface NSFetchedResultsDataSource : TTSectionedDataSource <NSFetchedResultsControllerDelegate>{
+
+	NSFetchedResultsController *_fetchedResultsController;
+	UITableView *_tableView;
+	NSEntityDescription *_entity;
+	NSString *_sortBy;
+	NSArray *_selectFields;
+	NSString *_sectionKey;
+	NSPredicate *_predicate;
+	NSTimer *_updateTimer;
+	NSInteger _fetchLimit;
+	NSArray *_relationshipsToFetch;
+	NSMutableArray *_delegates;
+	
+	BOOL _isLoading;
+	BOOL _isOutdated;
+    BOOL _tableIsUpdating;
+    NSUInteger sectionInsertCount;
+}
+
+@property (nonatomic, assign) BOOL isOutdated;
+@property (nonatomic, retain) NSMutableArray *delegates;
+@property (nonatomic, assign) BOOL isLoading;
+@property (nonatomic, retain) NSArray *relationshipsToFetch;
+@property (nonatomic, assign) NSInteger fetchLimit;
+@property (nonatomic, retain) NSTimer *updateTimer;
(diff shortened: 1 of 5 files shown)


jwang:
Note: This is the start of Three20 Core Data support.

MattNewberry:
It'd be interesting to look into my ObjectiveRecord library for possible integration, as it plays a big role in making Core Data easier.

For example, if you have an entity called Events, you can simply:

[Event pull:{completionBlock} fail:{failureBlock}];

Or, if you have an instance of an event...

Event *event = [Event blank];
[event fetch]; 

to update that object with remote data.

Anyhow, works VERY well for us

jwang:
yea, I would like to pull that in as well one day. I've used [this](https://github.com/yfactorial/objectiveresource) in the past and it's given me hell a few times.

MattNewberry:
I've been pulling my hair out on this for the last few weeks, but seem to be making progress.

In short, it looks like we need to have a new TTTableViewCoreDataController subclass, as the [self invalidateView]s that are called every time an object is updated is causing all sorts of problems.

By setting a _isLoading flag when the model starts and ends, and not letting [self refreshView] be called until then, it fixes the problem.

jwang:
Reading my mind! That's exactly what I'm working on creating. The part that I'm being careful and checking on, is making sure that any items that currently subclass TTTableViewDataSource can use either or. Such as TTThumbsDataSource, TTSectionedDataSource, and TTListDataSource

MattNewberry:
Makes sense... My version is hacky as of now, but it looks like its finally working.

Another caveat... When deleting a row using CoreData, layoutSubviews is being called on the cells again. When called, the object associated to that cell is now a fault because it's been deleted, and unless I explicitly do a check for [object isFault] and then return from layoutSubviews if it is, the app crashes as the object its trying to use for cell information doesn't exist.

MattNewberry:
Ping me on AIM ineedwords80days or GTalk matt@jadedpixel.com if I can be of any help... I've been fighting this integration the last few months and would be happy to lend my insight.

MattNewberry:
As I dig deeper, it looks like there's something going on still with the refreshView... 



Insertions are fine, however deleting is where the problem lays. 

When I trace the dataSource for numRows and sections, its as expected - 1 Section, 0 Rows

However, as you can see, the tableView doesn't completely refresh. Instead, three20 is covering that with the empty view, yet the table itself still has this errant view.

Any ideas?

MattNewberry:
Err... image link didnt work, link to image is http://cl.ly/3p3e453I3u3M0x301h18

MattNewberry:
I believe I finally figured out the issues... contact me for the solution unless you're far enough along with your own.

jwang:
Hey Matt, sending you message.

MattNewberry:
Just updated the branch with a very simple controller change... This can likely be refractored into the model.

The reason it's separate its because in my use-case, I load my local CoreData assets, then load the remote assets to update the local store.

Given that, I don't want the tableView to update to the Loading view when its already loaded local assets. By using a loading glag in the controller, we're able to have separate loading checks, one for remote and one for local. 

This simple change + my ActiveRecord port is working like a charm now.

jwang:
Thanks Matt. Just saw this, will take a look. :)

PR #499, opened 2011-03-28
Replaced NSLog with TTDINFO in CSSStyle

Hout:
Replaced NSLog with TTDINFO to allow for log entries to be numbed when not in DEBUG mode.


jwang:
thanks. not sure how that NSLog sneaked in.

PR #511, opened 2011-04-08
Fix TTURLRequest when POSTing files

rogchap:
1. I've added support for naming the file parameter name separately from the filename in a generatePostBody. 
2. I've also removed the first beginLine in generatePostBody as this was returning a HTTP error 400, and I could not see a need for it being there.

### DIFF
--- a/src/Three20Network/Headers/TTURLRequest.h
+++ b/src/Three20Network/Headers/TTURLRequest.h
@@ -274,7 +274,8 @@ extern const NSTimeInterval TTURLRequestUseQueueTimeout;
 /**
  * Add a file whose data will be posted.
  */
-- (void)addFile:(NSData*)data mimeType:(NSString*)mimeType fileName:(NSString*)fileName;
+- (void)addFile:(NSData*)data forKey:(NSString*)key mimeType:(NSString*)mimeType
+       fileName:(NSString*)fileName;
 
 /**
  * Attempt to send a request.
--- a/src/Three20Network/Headers/TTURLRequest.h
+++ b/src/Three20Network/Headers/TTURLRequest.h
@@ -273,8 +273,20 @@ extern const NSTimeInterval TTURLRequestUseQueueTimeout;
 
 /**
  * Add a file whose data will be posted.
+ * @param data the file data to be posted
+ * @param mimeType the mime type of the file
+ * @param fileName the file name
  */
-- (void)addFile:(NSData*)data forKey:(NSString*)key mimeType:(NSString*)mimeType
+- (void)addFile:(NSData*)data mimeType:(NSString*)mimeType fileName:(NSString*)fileName;
+
+/**
+ * Add a file whose data will be posted.
+ * @param data the file data to be posted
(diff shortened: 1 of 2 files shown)


rogchap:
FYI....don't know why I typed "port" when I really mean "post" :)

rogchap:
I've split out the new functionality so this should make sure it doesn't break current implementation.

RafaelAlba:
I tried using this patch in my fork of three20 and it seems to break all requests. The removal of this line seems to be the culprit(adding it back in there fixed the issue):
-    [body appendData:[[NSString stringWithFormat:@"--%@\r\n", kStringBoundary]  
-       dataUsingEncoding:NSUTF8StringEncoding]];

hongrich:
I believe the removal of those lines are what fixes the issues when POSTing files. I have not tried this patch directly, but I did remove those 2 lines just to fix the problem with POSTing files.

Could you try just remove those 2 lines but not use the rest of the patch to see if things still work?

rogchap:
@RafaelAlba What errors are you getting? I'm using this patch in all my applications that use both GET and POST requests with no issues.

RafaelAlba:
I removed those lines and the server fails with a 500 error, but with them there everything works fine. I'll have to check the exact error on the server tomorrow when I'm back in the office.

The server is running rails 3.x.

rogchap:
Are you getting the 500 on GET requests or is it only POST requests? What web server are you using to run Rails? Apache or Nginx (with Passenger)?

RafaelAlba:
Sorry about the long delay. Checking into this again. We're using nginx with passenger. The error in the logs is:

EOFError (bad content body).

It's happening on POSTs not sure about GETs just yet.

RafaelAlba:
Ok, only fails on POST requests

rogchap:
I've tested this and it seems that it is web server specific: 

When testing with IIS it fails with the extra boundary and succeeds without it, and Nginx seems to be the opposite: ie falis without the extra boundary and succeeds with the boundary line.

I need to understand more about post data / multipart forms / boundary to figure this one out.

jwang:
@rogchap can you update to development branch and resolve the conflicts accordingly? Also, not sure if and how you can deal with @RafaelAlba's issue of failing POST.

rogchap:
I'll update.

Not sure what to do on the POST problem. I've spent quite a lot of time on comparing "post"s from regular website form vs three20 (or any iOS project) using web proxies to check on the requests etc. I'm sure the extra boundary is wrong.

jwang:
Hey @rogchap - I'm going to push this off for next release. I want to test it a bit more as I'm concerned about it breaking existing implemention, though I think you already addressed that. I have 2 more items I want to push out today and tomorrow then will release. I'll put this into next week's release.

hongrich:
The boundary problem seem to have already been fixed in this commit abd35ca04a0c9e4f95bf9c22c55ee191ae086316

jwang:
Thanks @hongrich :)

jamesgpearce:

PR #519, opened 2011-04-12
& not url encoded

wwk:


tonklon:
This is not the only urlencoding problem in this method. The keys for example are not encoded at all. Please see related pull request https://github.com/facebook/three20/pull/387.

I think the urlencoding should be solved in a new method, taking the string encoding into account. IMHO this method should be left untouched to retain compatibillity with existing code.

PR #543, opened 2011-05-03
https links

mmorearty:
In styled text, automatically recognize https links in addition to http links.  For example, if styled text is this string, both URLs will be turned into hyperlinks:

```
This links to both https://www.google.com and http://www.google.com
```


jwang:
There's a regex way of doing this with [NSRegularExpressionSearch](http://developer.apple.com/library/ios/#documentation/StringsTextFonts/Conceptual/TextAndWebiPhoneOS/CustomTextProcessing/CustomTextProcessing.html)?

mmorearty:
That's true, but that is iOS 3.2/4.x only.

scashin133:
I had submitted a patch for this back in march as well: https://github.com/facebook/three20/pull/462

PR #548, opened 2011-05-05
[NETWORK] TTURLCaches loadImageFromDocuments Method not loading Retina Images

martinjuhasz:
fixed TTURLCache causing loadImageFromDocuments not returning Retina Images.

Tried to add TTLauncherItems with Images form the Documents Directory. Ended up with this solution


jwang:
Just a note, this doesn't quite resolve it fully. There's another issue here that I need to relate this to that is trying to solve a global issue with retina images not working in many situations. Another example is the Thumbnails in the photo gallery view.

jwang:
@martinjuhasz - I'm going to close this as it's targeting the master branch. I've seen this same request a few times, and many others have noted that it's not fully resolving for network items, but it does for local items. That's why it's working for the Documents Directory, but not for images coming down from Flickr and such. Would be great to find a solution for network items but that's more complicated.

PR #552, opened 2011-05-10
[UICommon] TTDeviceOrientationIsPortrait() and TTDeviceOrientationIsLandscape() helper functions

aporat:
Hi!

Two helper functions which saved me some code lines in my UIViews.

These two functions follow the naming of apple's UIInterfaceOrientationIsLandscape & UIInterfaceOrientationIsPortrait functions, which are a little annoying to use, as they must accept existing orientation.

These two new functions use TTDeviceOrientation() to easily check if the device is either in landscape or portrait.


aporat:
hum. something went wrong here

PR #556, opened 2011-05-13
Fix speechbubble

Hout:
Enabled arrows on left and right sides. Disabled arrows to be placed on corners; that looks ugly.

Three20 was unable to position the arrows on the speechbubbles on the side edges. This is corrected now.
In addition, positioning the arrow at 45 degrees would overlap the rounded corner. The arrow will now be placed right next to the rounded corner.

This has been mentioned on the forums in the past but proper code was never submitted.


jwang:
Thanks. can you provide a simple test case using the TTCatalog source as an example to verify?

Hout:
Done and it was a good idea as I identified a bug and fixed it ;-)
I included a few speech bubbles in TTCatalog with various pointers.

2011/5/29 jwang <
reply@reply.github.com>

PR #558, opened 2011-05-14
Add support for drawing the left and right edge of the speech bubble shape (TTSpeechBubbleShape).

buddydvd:
The current version only supports drawing edge on the top side and bottom side of the speech bubble. This change set adds support for drawing the edge on the left and right side of the speech bubble.

### DIFF
--- a/src/Three20Style/Sources/TTSpeechBubbleShape.m
+++ b/src/Three20Style/Sources/TTSpeechBubbleShape.m
@@ -134,6 +134,14 @@ static const CGFloat kInsetWidth = 5;
     CGPathMoveToPoint(path, nil, fw, RD(_radius));
   }
 
+  if (_pointLocation > 135 && _pointLocation < 225) {
+    CGFloat pw = _pointAngle >= 90 && _pointAngle < 270 ? _pointSize.width : -_pointSize.width;
+    CGFloat pointY = ((_pointLocation-135)/90)*fh;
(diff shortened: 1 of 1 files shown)


Hout:
:-) Looks like we have tackled the same issue here.
https://github.com/facebook/three20/pull/556

buddydvd:
What a coincidence. I will close this pull request then. It seems like the reason why the author didn't implement left/right edge drawing logic is because of the drawing-edge-near-the-corner problem (which your patch addresses) and lacking the ability to shift the edge vertically by a constant value (important for rendering bubbles with different heights). Also, with the way how the code is structured, it seems like the author intended to use _pointSize and _pointAngle to define the shape of the edge (non-isosceles triangles). The current implementation seems incomplete (the value of _pointAngle isn't really used in the code) and suffers some limitations (unable to create an obtuse triangle).

Hout:
Hmmm not entirely since the original triangles would be drawn on the corners too. However the whole idea of implementing a location based on an angle is simple but lacks precision. On a rectangular shape the _real_ 46 degree location varies to be located on the left and top sides depending on the rectangle's dimensions.

Implementation of an obtuse triangle crossed my mind but that would indeed collide with the functional suggestion of the interface.

I spent some hours last Friday night to investigate pointing in the right angle with a curved line alike the text messaging app. I learned a lot of things about arcs and curves but without tangible results.

For now this is good enough for me. I will re-investigate later.

PR #563, opened 2011-05-24
Restore views after memory warning: loading view, empty view, error view, table banner

mmorearty:
When iOS issues a memory warning, that causes several problems in Three20 for any views that are offscreen at the time of the memory warning, if those views are brought back onscreen later:
- If a TTModelViewController subclass is showing an "empty view", that empty view will be gone.
- Same for an "error view"
- Same for a "loading view"
- If a TTTableViewController subclass is showing a banner, that banner will be gone.

To reproduce this, download https://github.com/mmorearty/Three20MemoryWarningBugs, and follow the instructions.

This pull request fixes the issues.

(I would have also tested the "overlay view" feature of TTTableViewController, but I couldn't figure out how to get it to work.)

### DIFF
--- a/src/Three20UI/Headers/TTTableViewController.h
+++ b/src/Three20UI/Headers/TTTableViewController.h
@@ -48,6 +48,8 @@
 
 /**
  * A view that is displayed as a banner at the bottom of the table view.
+ *
+ * This property should be set within viewDidLoad.
  */
 @property (nonatomic, retain) UIView* tableBannerView;
 
@@ -102,6 +104,8 @@
 
 /**
  * Sets the view that is displayed at the bottom of the table view with an optional animation.
+ *
+ * This method should be called within viewDidLoad.
  */
 - (void)setTableBannerView:(UIView*)tableBannerView animated:(BOOL)animated;
 
--- a/src/Three20UI/Headers/TTTableViewController.h
+++ b/src/Three20UI/Headers/TTTableViewController.h
@@ -48,8 +48,6 @@
 
 /**
  * A view that is displayed as a banner at the bottom of the table view.
- *
- * This property should be set within viewDidLoad.
  */
 @property (nonatomic, retain) UIView* tableBannerView;
 
@@ -104,8 +102,6 @@
 
 /**
  * Sets the view that is displayed at the bottom of the table view with an optional animation.
- *
- * This method should be called within viewDidLoad.
  */
 - (void)setTableBannerView:(UIView*)tableBannerView animated:(BOOL)animated;
 
--- a/src/Three20UI/Sources/TTModelViewController.m
+++ b/src/Three20UI/Sources/TTModelViewController.m
@@ -208,6 +208,7 @@
 - (void)didReceiveMemoryWarning {
   if (_hasViewAppeared && !_isViewAppearing) {
     [super didReceiveMemoryWarning];
+    [self resetViewStates];
     [self refresh];
 
   } else {
--- a/src/Three20UI/Sources/TTTableViewController.m
+++ b/src/Three20UI/Sources/TTTableViewController.m
@@ -225,6 +225,15 @@
 - (void)loadView {
   [super loadView];
   self.tableView;
+
+  // If this view was unloaded and is now being reloaded, and it was previously
+  // showing a table banner, then redisplay that banner now.
+  if (_tableBannerView) {
+    UIView* savedTableBannerView = [_tableBannerView retain];
+    [self setTableBannerView:nil animated:NO];
+    [self setTableBannerView:savedTableBannerView animated:NO];
+    [savedTableBannerView release];
+  }
 }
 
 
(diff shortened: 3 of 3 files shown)


aporat:
I had some memory warning issues with the camera picker controller as well.

Try pushing the camera picker. In iphone 4, it will trigger a memory warning in many cases. after these warning are triggered, the main view (or entire controller - not sure) behind it get released resulting a blank white screen. To overcome this bug, I save the controller as a parameter in the app delegate.

Does this bug fix cover that bug as well?

mmorearty:
aporat: I'm not really sure.  I sort of doubt it.  Here's how the code works, and what my change does:

If a `TTModelViewController` (or any subclass of that, including `TTTableViewController`) starts loading the model, it call `[self showLoading:YES]`, to show a "Loading..." screen.  In the `TTTableViewController` subclass, the `showLoading` function creates a new view and displays it.

Later, when the model is done loading, `TTModelViewController` decides what to do next.  If the model encounters an error, then `TTModelViewController` calls `[self showError:YES]`.  Or if the model is empty, then it calls `[self showEmpty:YES]`.  Or in the normal success case, it calls `[self showModel:YES]`.

`TTTableViewController` handles `showError` and `showEmpty` in a similar way to how it handles `showLoading`: It creates another view and displays it.

But the problem comes in when a memory warning is issued.  The state of the `TTModelViewController` ends up getting confused.  E.g. suppose it previously called `[self showEmpty]` -- so it thinks it is now showing the "empty" page (and it is).  But then, while that view is offscreen, a memory warning comes in; so iOS "unloads" the view.  `-[TTTableViewController viewDidUnload]` gets called, and it destroys all the views that are associated with the empty view, error view, and loading view.  But the base class `TTModelViewController` thinks that (in my example) the "empty" view is still being displayed.

My fix is to change the `TTModelViewController` so that when a memory warning occurs, it explicitly resets its state.  That way, if the view gets loaded again later, `TTModelViewController` will know what's going on and will be able to restore its state.

The "banner" case is similar, although some of the details are different.

So...

As for the case you described, it isn't really clear to me what's going on.  When the camera picker starts (and loads its own modal view controller), that certainly shouldn't cause any other view controllers to be completely destroyed when a memory warning occurs.  It's odd to me that just saving your controller in your app delegate would fix this.  All that does is increment the refcount of the controller, thus preventing it from being destroyed; but that shouldn't be necessary.

jwang:
@aporat can you confirm your issue was solved? @mmorearty if you can verify that as well, it will help a lot! I'm ready to merge this as it's highly requested. :)

mmorearty:
@jwang Verified.

aporat:
@jwang sorry, my issue was actually unrelated to this patch, so i can't verify if this bug was fixed.

jwang:
Ah sorry @aporat I was trying to ask @buddydvd to confirm. Looking at multiple issues after another. :)

buddydvd:
@mmorearty: Can you please review the pull request I sent to your patch branch? The commit I did to change TTTableViewController.h was to support reverting changes to TTTableViewController.m as we discussed earlier.

jwang:
@buddydvd is it a separate request than the one I see on here from June 3rd?

buddydvd:
@jwang: That is a separate request. The commit mmorearty merged into this pull request was missing a second part. I submitted the second part for mmorearty to merge into this pull request and am waiting for his response.

jwang:
@buddydvd ah understood. Will wait for @mmorearty to get that into the request. Thanks! :)

jwang:
Three20 1.0.6 has been released. I'm going to add this into 1.0.7 for next week pending @mmorearty's review of @buddydvd's pull request for this change.

mmorearty:
I reviewed @buddydvd's pull request for this change (which was here: https://github.com/mmorearty/three20/pull/2 ), and decided not to take it.  It's a tough call, in my opinion.  Here's what I wrote on that page:

> @buddydvd, thanks for the pull request; but after thinking about it, I don't want to merge this pull request into my fork. I'm torn about it, but I think leaving the code the way it is is better.
> 
> If we make this change, and rely on the comment on the header file in order for people to avoid a memory leak, then essentially we are changing the contract for tableBannerView: Previously, there was no expectation that tableBannerView be set from viewDidLoad, and now there is. And what this means is, code that existed before now, and had this bug (the table banner view would suddenly disappear whenever a memory warning occurred), will still have the bug.
> 
> I prefer to change the code so that it fixes the bug while still adhering to the previous contract (that is, that the caller can set tableBannerView any time they want).

buddydvd:
@mmorearty: Agreed. Also, can you undo the commit I did for this pull request? That commit was intended to clarify when setTableBannerView should be called. Since you didn't accept my pull request's second commit, the first commit should be undone.

mmorearty:
@buddydvd ok, done.  @jwang this pull request is ready for merging.

PR #565, opened 2011-05-26
Fix tableView backgroundColor

ralfvdz:
UITableView backgroundColor always gray on iPad. Since SDK 3.2 one needs to initialize the backgroundView before setting the backgroundColor. See also: http://stackoverflow.com/questions/2688007/uitableview-backgroundcolor-always-gray-on-ipad

### DIFF
--- a/src/Three20UI/Sources/TTTableViewController.m
+++ b/src/Three20UI/Sources/TTTableViewController.m
@@ -643,6 +643,11 @@
     ? TTSTYLEVAR(tableGroupedBackgroundColor)
     : TTSTYLEVAR(tablePlainBackgroundColor);
     if (backgroundColor) {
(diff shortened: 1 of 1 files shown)


jwang:
interesting. Didn't know anyone was using the ipad-navigator branch. pretty sweet fix. Thanks!

ralfvdz:
No problem! Yes, I'm going to use this branch (as long as it's not in the master) on an iPad project :) I don't yet know for sure whether this problem also exists on the iPhone platform, so I put the pull request on this branch. 
I'm not 100% sure, but I guess the line "[_tableView setBackgroundView:nil];" is not required, because "[_tableView setBackgroundView:[[[UIView alloc] init] autorelease]];" will most probably release the old (gray) backgroundView and set the new one.

PR #581, opened 2011-06-21
Three20 CSS Extension New Design and TTImageViewDelegate improvements.

seqoy:
This update change the CSS Extension design adding an model object (TTCSSRuleSet) to each Rule Set parsed. This update keeps full compatibility with old versions but simplify access the CSS data like this:

// Read the color from the CSS 'headerFont'.
UIColor *headerFontColor = TTCSS( @"headerFont", color );

or

// Read the color rule set 'headerFont'.
TTCSSRuleSet *headerFont = TTCSSRuleSet( @"headerFont" );

UIColor *headerFontColor = headerFont.color;

Also introduces the TTCSSApplyProtocol, a common interface to classes that style itself using CSS readed properties. The first two UIKit classes to implement this protocol are UIView and UILabel.

### DIFF
--- a/src/Three20UI/Sources/TTPostController.m
+++ b/src/Three20UI/Sources/TTPostController.m
@@ -518,7 +518,7 @@ static const CGFloat kMarginY = 6;
 
 ///////////////////////////////////////////////////////////////////////////////////////////////////
 - (void)cancel {
-  if (!TTIsStringWithAnyText(_textView.text)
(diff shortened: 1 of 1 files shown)


seqoy:
Pull request pack. This pull also includes:

TTImageView Delegate improvement
- Give the user the possibility to configure the Requester before TTImageView make the call to the server.

jwang:
target development branch

PR #583, opened 2011-06-22
Don't scroll during update in TTTableViewController.

bmeurer:
In -model:didInsertObject:atIndexPath: of TTTableViewController, the
-scrollToRowAtIndexPath:atScrollPosition:animated: message was send
to the tableView whenever a new row was inserted. This is not going
to work if the new row was inserted as part of a larger change, i.e.
if -beginUpdates was called on the tableView and some new sections
were inserted prior to inserting the row.

In addition, scrolling to arbitrary rows without explicit control
from the App developer isn't good practice either, so just ditch
the scroll code all together.


diederich:
+1 on this from my side.

I've removed this since quite a while in my code, too. 
IIRC it could also happen that during batched updates from CoreData the TableViewController scrolled to rows which don't exist yet.

PR #584, opened 2011-06-23
TTLauncherView: Update pager's currentPage value if scrollToItem gets called.

alexleutgoeb:
The pager's currentPage value is not updated if scrollToItem:animated: scrolls to another page. A simple additional line fixes that.

### DIFF
--- a/src/Three20UI/Sources/TTLauncherView.m
+++ b/src/Three20UI/Sources/TTLauncherView.m
@@ -917,6 +917,7 @@ static const NSInteger kDefaultColumnCount = 3;
     NSUInteger page = [path indexAtPosition:0];
     CGFloat x = page * _scrollView.width;
     [_scrollView setContentOffset:CGPointMake(x, 0) animated:animated];
+    [self updatePagerWithContentOffset:CGPointMake(x, 0)];
   }
 }
 
--- a/src/Three20UI/Sources/TTLauncherView.m
+++ b/src/Three20UI/Sources/TTLauncherView.m
@@ -740,6 +740,12 @@ static const NSInteger kDefaultColumnCount = 3;
 
 
 ///////////////////////////////////////////////////////////////////////////////////////////////////
+- (void)scrollViewDidEndScrollingAnimation:(UIScrollView *)scrollView {
+  [self updatePagerWithContentOffset:_scrollView.contentOffset];
+}
+
+
+///////////////////////////////////////////////////////////////////////////////////////////////////
 ///////////////////////////////////////////////////////////////////////////////////////////////////
 #pragma mark -
 #pragma mark UIPageControlDelegate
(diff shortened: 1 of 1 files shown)


diederich:
Hi Alex,

thanks for looking into this.
I just saw the the LauncherView is supposed to update the currentPage when the scrollview ends scrolling:
https://github.com/alexleutgoeb/three20/blob/272717d2848dbf490c3bc0c7718ea54f5897d24d/src/Three20UI/Sources/TTLauncherView.m#L738 which seems to be consistent to the iPhone's home screen.

Just a guess, but are you calling this method with animated set to NO? Then the delegate might not be called and the pageIndicator isn't updated. I guess the fix for this, without changing current behavior might be to only call 

```
[self updatePagerWithContentOffset:CGPointMake(x, 0)];
```

when animated equals NO?

alexleutgoeb:
Hey Stephan,

I searched the header files (unfortunately the doc is not very descriptive) and found these hints:

```
- (void)scrollViewWillBeginDecelerating:(UIScrollView *)scrollView;   // called on finger up as we are moving
- (void)scrollViewDidEndDecelerating:(UIScrollView *)scrollView;      // called when scroll view grinds to a halt
- (void)scrollViewDidEndScrollingAnimation:(UIScrollView *)scrollView; // called when setContentOffset/scrollRectVisible:animated: finishes. not called if not animating
```

So as far as I've debugged the code, the `scrollViewDidEndDecelerating:` delegate method gets only called if the user triggers the scroll event, and not `setContentOffset:animated:` (the value of animated: makes no difference at all).

For the latter it would probably be better to implement the 3rd delegate method in place of my initial commit (though the problem with `animated:NO` would still persist)?

alexleutgoeb:
Stephan,

did have a closer look into this again, to finally answer your question more precise:

The problem is that the delegate method `scrollViewDidEndDecelerating:` gets only called if the scrolling was triggered by the user's finger. So the current implementation lacks of the page update on programmatically-triggered scrolling.

In order to to be consistent to the iPhone's home screen I would suggest two changes in code:
1. Implement the additional `scrollViewDidEndScrollingAnimation:`delegate method (body like the existing `scrollViewDidEndDecelerating`) for scrolling events with `animated:YES`
2. As you suggested, check for `animated == NO` in my first commit  for `animated:NO` calls

Hope you agree with that, will update the pull request with these changes.

diederich:
Hi Alex,

sorry, I pretty much forgot about this.
The last solution looks good to me. Thanks for looking into this!

 +1 for merging (@jwang)

diederich:
Just for reference (my link list):
- http://stackoverflow.com/questions/3034200/scrollviewdidendscrollinganimation
- http://stackoverflow.com/questions/993280/how-to-detect-when-a-uiscrollview-has-finished-scrolling
- http://developer.apple.com/library/ios/#DOCUMENTATION/UIKit/Reference/UIScrollViewDelegate_Protocol/Reference/UIScrollViewDelegate.html

jwang:
@alexleutgoeb - I'll be pulling this in later today, but just for future reference, make sure to set the commit range's base to `development` branch. It's currently targeting the `master` branch.

alexleutgoeb:
..and sry for the wrong commit range, was using github's "fork'n'edit" feature and forgot to check the destination at all.

jwang:
closed 47f665020e20f7b7f1b0b1256570593ecaf7e154

PR #585, opened 2011-06-24
Added a check to see if there are any "Copy Headers" phase files 

sumchattering:
If there is no "Copy Headers" build phase then the lint script crashes as it tries to concatenate NoneType.


sumchattering:
Merged with pull request #586

PR #315, opened 2010-11-03
Small change

clozach:
Hey Jeff,

Sorry for the delay in getting back to you. Congrats on what's looking like an awesome bugweek so far.

Anyway, here's the only real work I did that's worth merging in before I delete my fork. To give you some context, I used three20 for a client whose app included hundreds of images stored directly in the app. To my surprise, performance with pre-loaded images was far worse than remote-loaded images. This fixes that...hopefully without breaking anything from someone else's perspective.

Cheers,
Chris

PR #587, opened 2011-06-27
Resubmitting pull request to development instead of master

clozach:
Hi John,

Sorry for the delayed response...here's my previous pull request (https://github.com/facebook/three20/pull/315) submitted to development instead of master. I've gone ahead and closed #315.

Cheers,
Chris

Fixed a problem that causes performance problems for scrollviews that use TTPhotoView with images stored on-device.


jwang:
thanks! will start getting it in there.

PR #593, opened 2011-07-10
Support sending application/x-www-form-urlencoded forms 

jdewind:
Currently the TTURLRequest always submits a POST using multipart/form content type in order to support data and images out of the box. However, there are times when this is not necessary (most of the time) and some APIs stipulate the "Content Type" for a particular request -- returning an error if the content type is not respected.

This pull request contains a small set of changes that allows a URL request to be configured with a non "multipart" post body. By default, the request is multipart in order to be inline with the current expectations of a POST under TTURLRequest  -- which is every post contains a multipart/form body.

For example,

  TTURLRequest *urlRequest = [[[TTURLRequest alloc] initWithURL:@"http://www.google.com" delegate:nil] autorelease];

  [urlRequest setHttpMethod:@"POST"];

  urlRequest.multiPartForm = NO;

I have tests for these changes. However, I wrote them using a different test framework. The gist can be found [here](https://gist.github.com/823018).


jwang:
ah. gah. try and change the commit range to target the development branch.

jdewind:
Alright, so I've merged the latest from master onto my development branch, resolved the conflicts, which appears to be trivial (spacing) and added those commits to this merge branch. Note, there is a revert of a commit for a network semaphore that is no longer apart of this pull request.

jdewind:
I get it now. My eyes were equating development with master. New pull request submitted.

jwang:
Hehe. @dewind sorry about that. I think the commit range isn't as clear to most. You still have some conflicts with the current RC that I just pushed out to the Google Group, let me know if you can get those resolved in the new pull request. Thanks!

PR #600, opened 2011-07-12
Fix bug where URLs were always parsed if line breaks were parsed

rainhead:
Presently, if you try to parse some text with parseLineBreaks set to YES, the value of parseURLs will be ignored, and all URLs will be parsed. This commit fixes that logic by abstracting the operation of parsing a line into its own method, which is always sensitive to the value of parseURLs.

### DIFF
--- a/src/Three20Style/Headers/TTStyledTextParser.h
+++ b/src/Three20Style/Headers/TTStyledTextParser.h
@@ -44,5 +44,6 @@
 
 - (void)parseXHTML:(NSString*)html;
 - (void)parseText:(NSString*)string;
+- (void)parseLine:(NSString *)string;
 
 @end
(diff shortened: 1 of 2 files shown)


jwang:
Can you send a new one with the 'development' branch as the commit range? Thanks!

rainhead:
Sure, I'll get that to you tomorrow.

On Monday, July 11, 2011, jwang
reply@reply.github.com
wrote:

rainhead:
This looks like it was already fixed in development by commit f4e406bb, back in December. Would be awesome if this could get merged with master.

jwang:
@rainhead - yup. The development branch is currently being tested for stability. Please try it out and see if everything is working for you. The release is coming this week. It'll be version 1.0.6 which is currently in the development branch as release candidate stage. Thanks! :)

rainhead:
OK, good to know. I'll develop against development for a while and rush any bugs / fixes back to you. Thanks for your responsiveness.

PR #608, opened 2011-07-17
[UI] iAd support for setTableBannerView

aporat:
removed disabling the user interaction to allow Apple's iAds to be displayed & removed unused banner timer from TTTableViewController.

This will allow to pass an ADBannerView instead of TTImageView to [self setTableBannerView:banner animated:YES] function

### DIFF
--- a/src/Three20UI/Headers/TTTableViewController.h
+++ b/src/Three20UI/Headers/TTTableViewController.h
@@ -28,8 +28,6 @@
   UIView*       _errorView;
   UIView*       _emptyView;
 
-  NSTimer*      _bannerTimer;
-
   UIView*           _menuView;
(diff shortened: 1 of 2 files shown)


jwang:
Hmm. Not sure why that banner timer was there for.

diederich:
looks good to me, +1 for merging.

aporat:
maybe the timer was used to auto hide the banner after a specific amount of time.
I did a file search, and this var isn't being used anywhere in framework.

jwang:
@aporat ya. I looked around but didn't find anything. I'm going to put this into 1.0.7 for this week's release. I'm pushing 1.0.6.2 for the TTWebController issue separately.

PR #611, opened 2011-07-18
[UI] showBars shouldn't hide/show the status bar if it's initially hidden from the application plist settings

aporat:
when the status bar is initially hidden from the application plist settings, UIViewController showBars function shouldn't hide/show the status bar.

Declaring UIStatusBarHidden to be TRUE in the application plist file means the application is in full screen mode without the standard status bar. The showBars tries to show/hide the status bar right now and it's creating unexpected results, as the view layout is set to be full screen.

I had many issues with the UINavigationBar overlapping the status bar, and this pull should fix it as well.


jwang:
cool. thanks! I'll put this into the 1.0.7 release for this week. :)

jwang:
closed 2370253fcf624126d98bae1464f2f0569b623e89

PR #615, opened 2011-07-21
Update Complier to LLVM GCC 4.2

rogchap:
Update the projects to use LLVM GCC 4.2 compiler which is the default compiler on MacOS X Lion. Also removed obsolete build settings.

### DIFF
--- a/samples/Core/TTCoreDemo/TTCoreDemo.xcodeproj/project.pbxproj
+++ b/samples/Core/TTCoreDemo/TTCoreDemo.xcodeproj/project.pbxproj
@@ -319,9 +319,9 @@
 			buildSettings = {
 				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Developer";
 				GCC_C_LANGUAGE_STANDARD = c99;
+				GCC_VERSION = com.apple.compilers.llvmgcc42;
 				GCC_WARN_ABOUT_RETURN_TYPE = YES;
 				GCC_WARN_UNUSED_VARIABLE = YES;
-				PREBINDING = NO;
 				SDKROOT = iphoneos;
 			};
 			name = Debug;
@@ -332,9 +332,9 @@
 			buildSettings = {
 				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Developer";
 				GCC_C_LANGUAGE_STANDARD = c99;
+				GCC_VERSION = com.apple.compilers.llvmgcc42;
 				GCC_WARN_ABOUT_RETURN_TYPE = YES;
 				GCC_WARN_UNUSED_VARIABLE = YES;
-				PREBINDING = NO;
 				SDKROOT = iphoneos;
 			};
 			name = Release;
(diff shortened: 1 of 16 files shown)


diederich:
Looks good to me. +1 for merging

jwang:
How does this impact iOS 3.0 support? We've been fighting that for a while and we still need to support iOS 3.0 with Three20. I'll load this into my iOS 3.0 device and see how it goes. Thanks!

jwang:
If I recall correctly, this is either the 2nd or 3rd time someone has asked for LLVM + GCC and we had a problem with backwards compatibility on iOS 3.0 that we support. There was at one time a request to go full LLVM which we rejected due to that. I'll play around with this a little bit more.

aporat:
I'm for it as well. 

last time i checked on my iOS 3.1 device with 1.0.5, the app was broken anyway because of the TTSplitViewController class

jwang:
We've had mixed results recently with TTSplitViewController. I've had success just by making it a weak link, but others haven't. Some just pull it out. I think one of the iPad support views will fix it, but I haven't made my way through it as it's a really large review at this time. I'll push it up for next week's priority though.

diederich:
see #625 for an alternate approach without the compiler change

rahimizad:
Hi JWang, Why would anyone would still be using iOS3.0 (in the right mind of course)? iOS4.0 is about to go obsolete too soon after iOS5 is released next month. Why are you crippling this super wonderful library for those who can't grow up and adopt iOS4?

coneybeare:
@rahimizad While I agree that iOS 3.x is pretty nonexistent anymore, iOS 4 is nowhere near obsolete, even when 5.0 comes out. Adoption rates of new major iOS versions tend to go 50% in the first month, then a much slower adoption rate after that. If three20 would drop 4.0 support, a major portion of customers would be left behind.

rahimizad:
Thanks JWang, I was just trying to make a point that it is time to move off iOS 3.0 worries. I totally agree with you on iOS4.

aporat:
i think the major reason to keep iOS 3.1 support is because the first generation iPhone devices can't upgrade to iOS 4. meaning if your iPhone app requires iOS 4+, you lose some potential clients. I still think three20 should dump iOS 3.1 support - it's just not worth spending time & effort supporting iOS 3.1.

jwang:
I'm not sure if anyone looks at the milestones, but I'm anticipating dropping iOS 3 support and switching to LLVM + GCC 4.2 in 1.5 and then in 2.0 going towards ARC support.

Before we can do that, I want to make sure there's some sort of "stability" in the pre-1.5 branches. I'm operating based on iPhone 3G and iPhone basically being abandoned by Apple per iOS5 only supporting the 3GS and above.

These are pretty big undertakings. At some point, I am anticipating a bunch of people to start either sticking with Three20 or attempting to move to Nimbus. Three20 remains as one of the currently most watched and forked for Objective-C.

There's been several discussions about things.

**start @jwang's personal option:** 
Right now it's too big of a swiss army knife as opposed to say a specialty knife like a butter knife, bread knife, and steak knife. Certain knives cut things better than others.  I prefer systems that do 1 thing well and leave the rest to other items to do better. It's all about being the glue that puts things together than trying to solve everyone's problem, then not solving anything at all. 
**end @jwang's insane personal option.**

Lastly, if you're wondering. Yes, sometimes, I be crazy distracted and all over place. Sorry about that. You can blame the :bear:

aporat:
#742 was merged into development last week, which resolved these issues

PR #621, opened 2011-07-27
Three20/fix linter warning

diederich:
fixes warning about trailing whitespace


diederich:
wrong base. closing...

PR #625, opened 2011-07-28
xcode project renewal

diederich:
this is pretty much the same as pull request #615 but without the compiler changes (which is set through xcconfig and shouldn't be part of the project)


diederich:
this has been done with #742 including changes to the demo projects

PR #626, opened 2011-07-28
update SBJson to 3.0.2

diederich:
updates SBJson from 3.0betaX to 3.0.2 (removes obsolete files)
see #597 and #618

PR #636, opened 2011-08-03
[extJSON] update SBJson to 3.0.2

aporat:
In my previous pull request, I relied on an old SBJson version which was bundled with the BSYahooFinance library, which was a beta version of SBJSON.

this pull request updates the library to 3.0.2 (latest stable). I have tested the twitter sample app and the BSYahooFinance library, and they both behave normally with this update (and without any compiler notices / warnings)


diederich:
Thanks aporat!
This obsoletes #626 (and reminded me of missing pieces in TTExtensionLoader.m)

+1 for merging

PR #637, opened 2011-08-04
Fixed over-release of JSON string in TTURLJSONResponse

ericrwolfe:
Local variable json was replaced by an autoreleased string, then
explicitly released, causing a crash when the main autorelease pool
drained.


diederich:
Good catch, thanks!

Mind basing the pull request onto development?

ericrwolfe:
See https://github.com/facebook/three20/pull/639

PR #644, opened 2011-08-12
Issue387 urlencoding

tonklon:
This addresses issue #387. I think that pull should be closed and not pulled.


jwang:
sweet! Thanks @tonklon :)

PR #646, opened 2011-08-12
Ttpickertextfield selection

umairakeel:
Allow for a TTPickerViewCell to be highlighted and have an orange style.

Allow for a select of  TTPickerViewCell to also be sent to a delegate


jwang:
orange?

umairakeel:
The reason for orange is that we are using the TTPickerTextField as part of a email send application through our service. There may be times when the email address is either bad ( bounce backs ) or the person has unsubscribed; so the idea is to visually let the user know by highlighting the entry who it is.

Red seemed a bit harsh!

jwang:
ah. That's cool. There's some clean up stuff I want to do in there. Hmm. Maybe we can start looking at making it use the stylesheets then. That's much more preferable.

umairakeel:
By stylesheets do you mean other than the TTDefaultStyleSheet.m ?

jwang:
Nope. I mean let's use and add to it. :) - Meaning creating your own off of it should allow you to override accordingly. So the default is orange but if someone wants to change it, they just inherit from it. It's more of adding directions about how to change it from orange to another color. It's a wiki item. :)

umairakeel:
Ah,  OK, well let me try to figure out how to do that myself (have not used custom stylesheets yet) - and I will update the Wiki!

umairakeel:
so looking to hack on this a bit more over the weekend. Can you point me towards any existing docs / examples on stylesheets cause I am not sure where to being. 

the current change basically takes the existing code and just added the _highlighted_ state.

jwang:
Sure. I'll try to get something over to you tonight. I have a wedding to go to tomorrow, but I'll be on IRC #three20 channel and on Skype: jwang392 if you need me on Sunday. :)

jamesgpearce:

PR #647, opened 2011-08-17
TTTableAutomaticMoreButton

cagataygurturk:
This implementation extends TTTableMoreButton and TTTableMoreButtonCell classes and creates a new More button for TTTableView but slightly different than the original one. It does not need the user to click onto button to open more items to the tableview, so it creates a endless list.

This implementation firstly discussed in this stack overflow question:

http://stackoverflow.com/questions/6319476/three20-tttableviewcontroller-load-more-automatically/6321400#6321400


cagataygurturk:
I know it infinitely stays in the loading mode when there is no more data in the model. This is controlled in the DataSource. If you detect there is no more data, you simply do not show this button.

jwang:
I'm curious as to how this is different than using TTTableFooterInfiniteScrollView instead? Is that not usable for some reason?

jwang:
Ah. FYI TTTableFooterInfiniteScrollView was add on February 04, 2011 via 4c96f31fa11c974dfa6cb1d33c0be16ed60c63a0 I'm not sure if TTCatalog has a sample of it. If it doesn't, we should add one.

cagataygurturk:
ups i'm not aware of existence of such a feature and my implementation was old than february. thanks btw, let me see how is your solution :)

TTCatalog doesn't have an example of it.

cagataygurturk:
Maybe also my cell can be pulled because it shares the same design with the TTTableMoreButton and it allows to write a "Loading" text etc. I'm not sure, that's your decision.

jwang:
@cagataygurturk - I feel like for some reason, even so, it's probably safer to use the TTTableFooterInfiniteScrollView and have the "Loading" text in the footer's view? Maybe I'm confused.

PR #649, opened 2011-08-17
added -navigator:didNotMatchURL to the TTNavigatorDelegate optional selec

levous:
added -navigator:didNotMatchURL to the TTNavigatorDelegate optional selectors so that the delegate can respond to a path that wasn't in the TTURLMap.  Delegate selector will be called when mapped viewcontroller is nil and delegate responds to the selector

Having a delegate method the indicates the ttNavigator failed to match would be incredibly useful.  We're designing an app that has multiple sub navigation semantics and wanted to centralize and test navigation.  This is proving almost impossible using the UIView dependent navigator container infrastructure.  

I've added this selector and messaged it from the TTBaseNavigator when the controller resolves nil

thanks!

### DIFF
--- a/src/Three20UINavigator/Headers/TTNavigatorDelegate.h
+++ b/src/Three20UINavigator/Headers/TTNavigatorDelegate.h
@@ -45,4 +45,12 @@
 - (void)navigator:(TTBaseNavigator*)navigator willOpenURL:(NSURL*)URL
  inViewController:(UIViewController*)controller;
 
+/**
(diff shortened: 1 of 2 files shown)


jwang:
Cool! Thanks @levous perhaps we can make the name a bit closer to `didFailOpenURL:(NSURL *)URL` instead? Also, watch out for the commit range when you do the pull request. By default it will target the master branch and we push everything into the development branch first.

jamesgpearce:

PR #650, opened 2011-08-18
Empty tttableviewheaders

tonklon:
The heights for grouped table headers are not calculated correctly.

Remember, that the TTTableViewDelegate is used for grouped tables, too.

### DIFF
--- a/src/Three20UI/Sources/TTTableViewDelegate.m
+++ b/src/Three20UI/Sources/TTTableViewDelegate.m
@@ -45,6 +45,9 @@
 
 static const CGFloat kEmptyHeaderHeight = 0.f;
 static const CGFloat kSectionHeaderHeight = 22.f;
+static const CGFloat kGroupedSectionHeaderHeight = 36.f;
+static const CGFloat kGroupedSectionFirstHeaderHeight = 36.f + 10.f;
+static const NSUInteger kFirstTableSection = 0;
 
 ///////////////////////////////////////////////////////////////////////////////////////////////////
 ///////////////////////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


src/Three20UI/Sources/TTTableViewDelegate.m
```diff
@@ -124,7 +127,17 @@ static const CGFloat kSectionHeaderHeight = 22.f;
     if (!title) {
       return kEmptyHeaderHeight;
     }
-    return kSectionHeaderHeight;
+
+    if (tableView.style == UITableViewStylePlain) {
+      return kSectionHeaderHeight;
+
+    } else {
+      if (section == kFirstTableSection) {
```
tonklon, line 135: The first section used to be 10 points higher. Optically all header are the same height, because all following header are preceded by a 10 point default footer from the section above.

If this distinction is not made all first sections of grouped tables will be visually shortened by 10 point.

src/Three20UI/Sources/TTTableViewDelegate.m
```diff
@@ -45,6 +45,9 @@
 
 static const CGFloat kEmptyHeaderHeight = 0.f;
 static const CGFloat kSectionHeaderHeight = 22.f;
+static const CGFloat kGroupedSectionHeaderHeight = 36.f;
+static const CGFloat kGroupedSectionFirstHeaderHeight = 36.f + 10.f;
```
tonklon, line 49: Xcode does not let me write `static const CGFloat kGroupedSectionFirstHeaderHeight = kGroupedSectionHeaderHeight + 10.f;`, which I would prefer. It says this was not a constant expression. ???

diederich, line 49: IIRC C can't do that. Need C++ for that ;)

PR #652, opened 2011-08-19
CGFloat Consistency

jwang:
It's a coding style thing. We should really be consistent on this and follow Apple Guidelines on it.


jwang:
Oops. My spiffy Icon.png for Git Tower snuck in there. I can remove it if it bugs people. :P

diederich:
hehe, thanks for the icon. Didn't know that.

And thanks for changing all the floats!

PR #654, opened 2011-08-19
Allow the labels for TTTabGrid to be visible

jwang:
In TTCatalog if you look at the TabGrid in the Tabs section, none of the labels are visible. By removing TTPartStyle from the default style sheet, the labels become visible. There is an underlying problem with TTPartStyle when it's sent a CGSizeZero TTImageStyle. It's causing weirdness in TTButton's `drawRect:` method to flip out when doing math on `imageBoxStyle.margin.right`
There seems to be very little point in having TTPartStyle in there for now as it's returning an empty image with nil default and nil image. It seems to be a high power user feature.

This will close out :octocat: #131 until someone has time to go into TTStyle and deal with the funky math.


jamesgpearce:

PR #656, opened 2011-08-20
Discovered a crash when parsing Google geocode JSON response. 

goof:
Fixing a bug scenario when we lose the reference of the original json string, if stringByTrimmingCharactersInSet returns with an autoreleased object.

http://maps.googleapis.com/maps/api/geocode/json?address=210%20Castro%20St,%20Mountain%20View&sensor=false


jwang:
This is actually fixed in the development branch that's almost ready for release. There's 1 more tweak I want to do in there, then it'll be ready.

goof:
great to know, thanks john!

PR #657, opened 2011-08-21
TTDevicePlatform() function

aporat:
A little useful function to get a human readable string of the device's platform


jwang:
sweet. Thanks!

jwang:
@aporat - Do you think we should make this be used in [TTAvailability](https://github.com/facebook/three20/blob/development/src/Three20Core/Headers/TTAvailability.h) by adding a iOS version?

aporat:
im not sure. 

i think TTAvailability is more about iOS version availability, while this new device model just returns the model name in a human readable string.

aporat:
oh, and i guess the function should be named TTDeviceModelName() - similar to [[UIDevice currentDevice] model] (which returns only the basic model name)

jwang:
that makes sense to me. :)

PR #670, opened 2011-09-06
iPad Split View updates

aporat:
some code cleanups i've made for a few of my three20 universal apps. 
- renamed left & right navigator to primary & secondary navigators
- fixed an issue with disappearing split button when changing the interface orientation


jwang:
I really wish we had better iPad support. This is really awesome @aporat! Let's slate this for a 1.1 release as it's going to break some people's apps. We need to add deprecation warnings since you changed from `left` and `right` to `primary` and `details`. You can either keep the left and right and have them call primary and details and add the deprecation warnings.

Or just break all hell and add in the changelog notes that it's being deprecated.

aporat:
yeah, it was really a challenge making a three20 universal app with both portrait & landscape orientations.

I'll try to keep working on these changes (and borrowing some code from https://github.com/facebook/three20/pull/533 as possible). This pull request stands by itself, and i think it can be releases as is (I have 3 iPad apps with these changes on the app store)

I think it's fine to break the current iPad support if it will be release it as a part of 1.1. Keeping the left & right old navigations will be messy at best..

hum, where's the change log anyway? never knew there's one :-)

jwang:
I'm still not sure what to do with #533. Maybe it'll be what goes into the 1.5 that starts dropping support. I do want to get working on that soon. Changelog is a file that doesn't currently exist in the main repo, but in my local one for now. I need to push that out. I want people to start using the Wiki here more too.

aporat:
well, someone will have to finish #533 before  :-) I tried testing it awhile ago, and it's still missing some key features (and it's more buggy than what three20 has now)

I was thinking about trying to integrate https://github.com/mattgemmell/MGSplitViewController into three20 as a replacement for the current TTSplitViewController. 

MGSplitViewController is much more powerful than the standard UISplitViewController with some extra cool enhancements.

jwang:
I love MGSplitViewController. I use it as well instead of the TTSplitViewController.

PR #673, opened 2011-09-06
New methods in TTCSSRuleSet to retrieve attributes for Core Text formatting.

seqoy:
- New method <code>paragraphAlign</code> - Return an <code>CTTextAlignment</code> based on the
  <code>text_align</code>CSS property.
- New method <code>attributedStringDictionary</code> -
  Return an Dictionary with formatted attributes in order to use with
  <code>NSAttributedString</code> objects.
- Now accept the <code>font-style</code> CSS property. 
- Implement the <code>NSCopying</code> protocol. 
- New <code>coreTextFont</code> method return an Core Text Formatted Font (<code>CTFontRef</code>) 
- <code>UIFont</code> are created using Core Text formatting. So you can use real <code>bold</code> and <code>italic</code> properties and the class will pick the
  correct System Font.


jwang:
Ooo. Love the NSCopying protocol implementation. Any ideas how the Core Text will work on older devices? Just a question, though we likely won't care too much in the near future.

seqoy:
Yes, <code>NSCopying</code> is cool. <b>Core Text</b> is available since <i>iOS 3.2</i>, isn't so new. I'll put some conditionals compiler and tests to disable this specific methods, and create <code>UIFont</code> using only <i>3.1</i> code. With that I think we're covered.

seqoy:
@jwang: Take a look now. I did separate Core Text methods under a Category. And add compiler conditions for the correct iOS.

jwang:
There's some conflicts with the development branch that need to be resolved.

seqoy:
Ok. I'll see that.

jamesgpearce:

PR #674, opened 2011-09-07
Set TTTableMessageItemCell height dynamically based on its content

cglong:
Right now, the height of `TTTableMessageItemCell` is hard-coded to be 90 pixels. This usually works, but in my case, I didn't want to display a caption. This led to 15 pixels of whitespace at the bottom of every cell.

This implementation assumes there's a title, like other item cells. It checks if there's a caption and/or text and increases appropriately. It also checks if an image is present and overrides the height if necessary.

I chose to multiply `kTableCellVPadding` by 1.5 rather than 2 to not break existing uses of this cell. I think it also "looks" better being 90 pixels high, as opposed to 95.


jamesgpearce:

PR #675, opened 2011-09-07
Added NSURLConnectionDelegate methods to TTRequestLoader to prevent Xcode warnings

dcai:
Added NSURLConnectionDelegate methods to TTRequestLoader to prevent Xcode warnings


jwang:
We're going to be fixing this in a separate way. Thanks!

PR #680, opened 2011-09-09
Fix for TTURLRequest / NSURLRequest immediate timeout issues

matej:
This fixes a bug where a timeout value of 0.0 seconds is used in a NSURLRequest created without a given TTURLRequest. 

This causes the NSURLRequest to timeout immediately and fail.
The TTURLRequest will always be nil when there are several requests in a TTReuestLoader.

This issue can easily be seen by taking a look at the relevant source code and caused a lot of failed request for me (loading a lot of images in a TTTableView). 

In `TTRequestLoader -connectToURL:` request is nil when there is more than one request in _requests:

``` objc
TTURLRequest* request = _requests.count == 1 ? [_requests objectAtIndex:0] : nil;
NSURLRequest* URLRequest = [_queue createNSURLRequest:request URL:URL];
```

In `TTURLRequestQueue -createNSURLRequest:URL:` the usedTimeout is set from the passed in request:

``` objc
NSTimeInterval usedTimeout = request.timeoutInterval;

if (usedTimeout < 0.0) {
  usedTimeout = self.defaultTimeout;
}
```

If `request == nil` usedTimeout becomes 0.0 and thus causes an immediate timeout. 

Regards,
Matej


PabloGS:
I can confirm this bug and fix.

jwang:
Thanks for verifying @PabloGS
@matej - can you resubmit with the commit range being the development branch? Thanks.

ccummiskey:
I ran into this problem too.  It occurs very rarely, but it is a pain to troubleshoot when it does.  The error occurs when you already have more then 5 simultaneous requests and then make multiple requests with the same cache key.  The better fix would be to change both occurrances of the following lines from TTRequestLoader.m: 

TTURLRequest\* request = _requests.count == 1 ? [_requests objectAtIndex:0] : nil;

to

TTURLRequest\* request = _requests.count >= 1 ? [_requests objectAtIndex:0] : nil;

This behaviour is more correct, and you won't lose your header dictionary, httptype, etc.

jwang:
@PabloGS  @matej - have you looked at the proposed fix from @ccummiskey and @scompt. It looks like that will be a more related fix?

tonklon:
@jwang I will look into this and find out which is the right fix. I have to admit, that I introduced this bug with 604a5faa71161652403a5e7aff62ffe84a70984f

tonklon:
merged into development

PR #683, opened 2011-09-13
fix accessory width in caption cells - closes #189

dwery:
fix for #189

PR #685, opened 2011-09-13
fix cell width when an URL is present

dwery:


phillipuniverse:
duplicate of #683

jwang:
closed per duplicate of #683

PR #686, opened 2011-09-14
TTURLNavigator minor compiler warning fix

aporat:
nothing interesting here, just a minor compiler warning i found while looking into a three20 issue with external URLs...


jwang:
sigh. Thanks!

PR #689, opened 2011-09-23
TTTableViewCell background color issue on iPad devices

aporat:
iPad table cell backgrounds aren't completely white, they have a little dark gray color background. This is an issue, as the background colors in the default style sheet are set as white. this creates a ugly white overlay if you use TTTableView text cells.

This change just change the background to clearColor, which blends nicely into the background (on both iPad & iPhone)


coredream:
I see this issue as well on the iPad, and this fixes it.  But note TTThumbsViewController ends up with a black background by default since it also uses TTSTYLEVAR(backgroundColor).

aporat:
Yeah, i had the same issue as well.

we can either solve it by creating new styles just for the tables, such as tableBackgroundColor, or by updating the existing code to make sure the text background layer are white. I rather go over the TTViewControllers updates to avoid adding new styles

comment?

coredream:
I agree, the latter is a better solution instead of creating new styles.

aporat:
the last commit fixes the issue with the background color in the photo controller, so it's okay to merge as far as I know..

colinyoung:
just a bump, i'm seeing this on the iPhone 5.0 simulator and device as well.  @aporat's fix worked perfectly to fix the problem.

aporat:
@diederich can you merge? it is a simple and important iPad fix

diederich:
While this might have quite an performance impact, I agree the default should be clearColor. 
Devs will find this issue during performance optimization anyways and can fix/change it then.

PR #692, opened 2011-09-26
Fix messagecontroller background

Hout:
Corrects the white background of field labels in TTMessageController.


jwang:
This seems to be coming up a lot more often now. #689 also deals with an issue of white background colors.

Hout:
Yep, but I want the background of the label to match the general background which in my case is not white.

PR #681, opened 2011-09-10
Support for sections with footer text in TTSectionedDataSource  

aporat:
This change adds the ability to have custom footer text in table view sections.

A `TTSectionedDataSource` section can be either a standard `NSString` or a `TTTableSection` object. The section object can be used to set both header & footer text (or any combination of the two). 

`TTTableSection` class has one helper function which is used create sections, similar to the behavior of `TTTableItem`:

`[TTTableSection sectionWithHeaderTitle:@"Header Text" footerTitle:@"Footer Text"]`

See the modified TTCatalog with some examples.

PR #693, opened 2011-09-26
[UI] Adding TTTableSection

aporat:
This change adds the ability to have custom footer text in table view sections.

A `TTSectionedDataSource` section can be either a standard `NSString` or a `TTTableSection` object. The section object can be used to set both header & footer text (or any combination of the two). 

`TTTableSection` class has one helper function which is used create sections, similar to the behavior of `TTTableItem`:

`[TTTableSection sectionWithHeaderTitle:@"Header Text" footerTitle:@"Footer Text"]`

See the modified TTCatalog with some examples.

### DIFF
--- a/samples/TTCatalog/Classes/TableItemTestController.m
+++ b/samples/TTCatalog/Classes/TableItemTestController.m
@@ -21,8 +21,8 @@ static NSString* kLoremIpsum = @"Lorem ipsum dolor sit amet, consectetur adipisi
     self.title = @"Table Items";
     self.variableHeightRows = YES;
 
-    // Uncomment this to see how the table looks with the grouped style
-    //self.tableViewStyle = UITableViewStyleGrouped;
+    // comment this to see how the table looks with the standard style
+    self.tableViewStyle = UITableViewStyleGrouped;
 
     // Uncomment this to see how the table cells look against a custom background color
     //self.tableView.backgroundColor = [UIColor yellowColor];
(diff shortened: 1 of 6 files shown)


jwang:
Note to self: Resubmit of #681 with conflicts resolved.

samples/TTCatalog/Classes/TableItemTestController.m
```diff
@@ -21,8 +21,8 @@ static NSString* kLoremIpsum = @"Lorem ipsum dolor sit amet, consectetur adipisi
     self.title = @"Table Items";
     self.variableHeightRows = YES;
 
-    // Uncomment this to see how the table looks with the grouped style
-    //self.tableViewStyle = UITableViewStyleGrouped;
+    // comment this to see how the table looks with the standard style
+    self.tableViewStyle = UITableViewStyleGrouped;
```
jwang, line 25: @aporat - do you think it's a better idea to have one that shows Grouped and one that shows Plain instead of asking people to comment out?

aporat:
yeah, but it means adding another test controller to the catalog sample app, which is already packed with many features.

Maybe a separate sample project just for TTTables showing off the main features of three20 tables?

aporat:
just a reminder that it's ready for merging :-)

kemenaran:
I'll be glad to have this merged: it just got the commits manually from the kosher-penguin repo, because I needed the feature.

From my point of view, it works flawlessly. This would be a nice addition to Three20's trunk.

Anyone to do the merging?

PR #695, opened 2011-10-03
iOS 3.1 compatibility changes

aporat:
This request advances three20 iOS 3.1 compatibility (I'm using it in few of my apps)
- Using TTUIPopoverControllerClass() function to create the UIPopController 
- Fixed a direct call to userInterfaceIdiom, which was changed to TTIsPad()

The only issue left with three20 and iOS 3.1 compatibility is the TTSplitViewController controller, which can be easily removed (or replaced)

### DIFF
--- a/src/Three20UI/Sources/UITableViewAdditions.m
+++ b/src/Three20UI/Sources/UITableViewAdditions.m
@@ -24,6 +24,7 @@
 
 // UICommon
 #import "Three20UICommon/UIWindowAdditions.h"
+#import "Three20UICommon/TTGlobalUICommon.h"
 
 
 ///////////////////////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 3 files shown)


jwang:
Ideally, the TTSplitViewController issue will go away soon with iOS 5 being released. In all likelihood I am anticipating Apple only really supporting iOS4 and 5 and dropping support for iOS3.x and below. As I understand it, there's a small subset that will be left out. Likely anyone with a 3GS that doesn't upgrade to iOS5 or the new iPhone. If there's anyone still running on a 3G at that point, it's probably not going to be many developers' target audience. This is good, I'm hoping to start dropping support slowly for older OS versions. I think I outlined them in milestones, but need to make it clearer in the Wiki.

aporat:
merged into development (1.0.11-dev)

PR #698, opened 2011-10-05
iOS 5 compatibility for dismissing modal view controllers

rehos:


saintnicster:
1.  Is it ok to add the iOS 5 stuff yet?  I thought we were technically still under the NDA... (first release as a developer, so I may just be out of the loop)
2.  Is this really needed?  dismissModalViewController is still working for me.  What is the benefit of using this instead?

PR #700, opened 2011-10-07
Preventing occasional crash in TTURLRequestQueue :: createNSURLRequest:URL:

ettore:
The crash happens because we're attempting to call NSURL :: URLWithString: with a request.urlPath == nil.  I am aware that TTURLRequestQueue checks for blank urlPath's before queueing the request, but somehow this situation happens anyway. (Maybe during a memory warning cleanup?)

So I added a 2-liner sanity check to improve stability: this simply cancels the request in the case described above.


jwang:
The lint should have thrown some warnings here. Also can you resubmit this with targeting the development branch instead of the master branch. Thanks!

jwang:
thanks @ettore for resubmitting. Note: see #712

PR #703, opened 2011-10-13
Found a bug in a call to the TTDWARNING method

marzapower:
I found this bug while mapping the navigator to a URL that generates no objects. The call to the TTDWARNING method, made inside the TTURLNavigationPattern class, has no NSString\* parameters, even if it uses a "%@" tag inside. 

Fixed the call to the TTDWARNING method extracting the URL parameter from the string

### DIFF
--- a/src/Three20UINavigator/Sources/TTURLNavigatorPattern.m
+++ b/src/Three20UINavigator/Sources/TTURLNavigatorPattern.m
@@ -487,7 +487,7 @@ static NSString* kUniversalURLPattern = @"*";
       returnValue = [self invoke:target withURL:URL query:query];
 
     } else {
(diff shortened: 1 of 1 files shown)


jwang:
can you resubmit this against the development branch instead of the master branch? Thanks!

marzapower:
Sure! Just give me the time to get the right branch.
I'm sorry for the inconvenient.

2011/10/13 John Wang <
reply@reply.github.com>

marzapower:
Never mind, in the development branch this bug has already been fixed.
Thank you and please forgive me!

2011/10/13 Ing. Daniele Di Bernardo daniele.dib@marzapower.com

PR #716, opened 2011-11-04
cache://

clarketus:
Hey guys

I just got an app rejected saying I was storing information in the /Documents folder that I should be storing in the /Library/Caches folder

From their rejection notice:
"For example, only content that the user creates using your app, e.g., documents, new files, edits, etc., may be stored in the /Documents directory - and backed up by iCloud. Other content that the user may use within the app cannot be stored in this directory; such content, e.g., preference files, database files, plists, etc., must be stored in the /Library/Caches directory."

So this is fine, I went to move the locations where my application stored files, but I then noticed I was displaying some of these files inside a TTPhotoViewController. I was previously using documents:// to access the files, and had no easy way to access them now they were in the /Library/Caches folder

So, I have added cache:// to the provided URL protocols. This could be a candidate for pulling into three20? If I should be doing this a different way, let me know.

Thanks!

### DIFF
--- a/src/Three20Core/Headers/TTGlobalCorePaths.h
+++ b/src/Three20Core/Headers/TTGlobalCorePaths.h
@@ -27,6 +27,11 @@ BOOL TTIsBundleURL(NSString* URL);
 BOOL TTIsDocumentsURL(NSString* URL);
 
 /**
+ * @return YES if the URL begins with "caches://"
+ */
+BOOL TTIsCachesURL(NSString* URL);
+
+/**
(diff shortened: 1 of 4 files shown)


jwang:
Yea. that new change is going to catch a lot of people now. Thanks will take a look at this. :)

clarketus:
Hey

Is there any chance this can get merged? Or should I close this.

Thanks

diederich:
Hi thanks for the pull request. There are some minor formatting changes in the diff, but I guess that's okay.
Merging...

PR #712, opened 2011-10-25
(dev) Preventing occasional crash in TTURLRequestQueue :: createNSURLRequest:URL:

ettore:
Resubmitting this pull request ( https://github.com/facebook/three20/pull/700 ) to development branch.

The crash happens because we're attempting to call NSURL :: URLWithString: with a request.urlPath == nil. I am aware that TTURLRequestQueue checks for blank urlPath's before queueing the request, but somehow this situation happens anyway. (Maybe during a memory warning cleanup?)

So I added a 2-liner sanity check to improve stability: this simply cancels the request in the case described above.

PR #717, opened 2011-11-05
Preventing occasional crash in TTURLRequestQueue :: createNSURLRequest:URL:

ettore:
Resubmitting pull request #700 (and #712) to development branch, removing unrelated changes.

The crash happens because we're attempting to call NSURL :: URLWithString: with a request.urlPath == nil. I am aware that TTURLRequestQueue checks for blank urlPath's before queueing the request, but somehow this situation happens anyway. (Maybe during a memory warning cleanup?)

So I added a 2-liner sanity check to improve stability: this simply cancels the request in the case described above.


jwang:
@ettore - This is a resubmit of #700 and #712?

ettore:
Yes: I committed too many unrelated changes to #712, so I closed that one and resubmitted just the file I intended originally. Sorry for the confusion.

PR #721, opened 2011-11-17
TTTabStrip updateOverflow not being called 

aporat:
closes #696. 

I had the same layout issue in my apps


jwang:
sweet. Thanks!

aporat:
@jwang, opps. looks like we forget to actually merge this pull request (so wasn't released in 1.0.10)

PR #725, opened 2011-11-26
Add information to TTLauncher when an item is moved. From where and to where

saimonx:
I have added a new method on the TTLauncherViewDelegate.

Also I have modified the TTLauncherView.h to add a variable, because the start position of the item is overwritted each time you move his position.

Also, the call to the delegate from the TTLauncherView.m and the set of the new variable


jamesgpearce:

PR #726, opened 2011-11-26
Warning suppression and .pbxproj files updated

Vaseltior:
I did some minor corrections over getters that where called that way : object.selector instead of [object selector]. So there is no annoying warnings for the one I found.

Next I corrected some mismatched types like some NSData that were affected to NSMutableData objects.... Bad bad bad :-)

I saw something strange in the prototype definition of a method in the navigator, so I did the correction. But the fact is that I'm not sure of it.

Also some warning between UIInterfaceOrientation and UIDeviceOrientation. We know that interface orientations are subset of device orientation. So I fixed that way.


src/Three20Network/Sources/TTRequestLoader.m
```diff
@@ -100,7 +100,9 @@ static const NSInteger kLoadMaxRetries = 2;
     // Strictly speaking, to be really conformant need to interpret %xx hex encoded entities.
     // The [NSString dataUsingEncoding] doesn't do that correctly, but most documents don't use that.
     // Skip for now.
-	_responseData = [[[dataSplit objectAtIndex:1] dataUsingEncoding:NSASCIIStringEncoding] retain];
```
Vaseltior, line 103: The given data is a NSData object while the _responseData is a NSMutableData object.
So using it that way could crash if you try to use a NSData as a NSMutableData

src/Three20Network/Sources/TTRequestLoader.m
```diff
@@ -140,6 +142,111 @@ static const NSInteger kLoadMaxRetries = 2;
   }
 }
 
+///////////////////////////////////////////////////////////////////////////////////////////////////
```
Vaseltior, line 145: I simply moved the NSURLConnectionDelegate methods so that they are known by the other methods as they are higher in the file. It remove some warnings

src/Three20UI/Sources/TTPhotoViewController.m
```diff
@@ -379,7 +379,9 @@ static const NSInteger kActivityLabelTag          = 96;
       // The photo source has a URL mapping in TTURLMap, so we use that to show the thumbs
       NSDictionary* query = [NSDictionary dictionaryWithObject:self forKey:@"delegate"];
       TTBaseNavigator* navigator = [TTBaseNavigator navigatorForView:self.view];
-      _thumbsController = [[navigator viewControllerForURL:URL query:query] retain];
```
Vaseltior, line 382: Mismatching object type

src/Three20UICommon/Sources/TTGlobalUICommon.m
```diff
@@ -113,7 +113,8 @@ BOOL TTIsPad() {
 
 ///////////////////////////////////////////////////////////////////////////////////////////////////
 UIDeviceOrientation TTDeviceOrientation() {
-  UIDeviceOrientation orient = [UIApplication sharedApplication].statusBarOrientation;
```
Vaseltior, line 116: statusBarOrientation is UIInterfaceOrientationType... UIDeviceOrientation is not :-)
But UIInterfaceOrientation is a subset of UIDeviceOrientation so... this new code works perfectly

src/Three20UINavigator/Sources/TTURLNavigatorPattern.m
```diff
@@ -487,7 +487,7 @@ static NSString* kUniversalURLPattern = @"*";
       returnValue = [self invoke:target withURL:URL query:query];
 
     } else {
-      TTDWARNING(@"No object created from URL:'%@' URL");
```
Vaseltior, line 490: Somebody forgot a comma there

sumchattering:
Or we can simply add the protocol NSURLConnectionDataDelegate to TTRequestLoader.Either way works.

aporat:
most of these issues have been resolved with #742 & #741

PR #730, opened 2011-12-01
Fixed incorrect control's width

tszming:
When the control return `YES` in the method `shouldConsiderControlIntrinsicSize`, e.g. `UISwitch`, besides the position `x` should be set, the width also need to be updated.

You didn't observe any problem in the existing code since only the `UISwitch` is the candidate - it's width is fixed anyway, but when you extend to more controls, you will see the problem.


aporat:
merged into development (1.0.11-dev)

PR #731, opened 2011-12-02
added a searchMessageView to TTSearchDisplayController

ghost:
This view (if present) is automatically "faded-in/out" on top of the private _dimmedView.
The view could contain some instructions, or images and allow the user to be notified before he starts to type something.


jamesgpearce:

PR #732, opened 2011-12-04
Immediate timeout fix

scompt:
This is the isolated fix from @ccummiskey in pull request #680. I'm not sure what the relation is to d2bf8d20 from @ettore. They both fix code around the same couple lines. If they're both trying to fix the same problem, I think this pull request is a better fix because it doesn't involve canceling the request.

### DIFF
--- a/src/Three20Network/Sources/TTRequestLoader.m
+++ b/src/Three20Network/Sources/TTRequestLoader.m
@@ -119,7 +119,7 @@ static const NSInteger kLoadMaxRetries = 2;
   }
   TTNetworkRequestStarted();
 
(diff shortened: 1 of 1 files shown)


tonklon:
merged into development

ettore:
I think my pull request d2bf8d2 fixes a different problem. My fix is not related to any timeout issue: it's simply preventing a crash caused by trying to create a NSURLRequest from a (non-nil) TTURLRequest whose urlPath property is nil.

tonklon:
I too think d2bf8d20ad391b4c05d0d95f4f5f472c968f2295 is a different problem.
Nevertheless d2bf8d20ad391b4c05d0d95f4f5f472c968f2295 was already pulled with request #700

PR #737, opened 2012-01-04
Appledoc API Documentation

aporat:
this pull request propose is to replace http://api.three20.info/ with an appledoc api docs hosted on github pages.

The docs.py script generates an appledoc doc set (which can be imported into xcode) and a html version of the api docs. the script also generates a gh-pages subfolder which can be pushed into the gh-pages branch, as a part of the release process.

To generate the API docs using appledoc, use the docs.py, for example:

`python src/scripts/docs.py --version 1.0.10-dev --generate`

I'm not that experienced with python scripting, so the script is pretty basic now. might be nice to have the gh-pages branch pushed automatically in the future.


tonklon:
I don't think it's a good idea to add the appledoc binary to the repository. People can get the doc feed from gh-pages. If they want to generate it themselves they could install appledoc. Also the template files are missing, so this setup is not working right now, because users that do not have the templates in ~/Library/Application Support/appledoc will not be able to generate documentation.

I would prefer to use an AppledocSettings.plist file to specify project-settings instead of a script. That would separate configuration from code.

aporat:
Ideally, the project should have script that builds the docs after each release, clone the github gh-pages and publish the new docs automatically. Currently, the script does the docs generation based on a version number and create the folder layout for gh-pages.

I couldn't get the latest appledoc to generate doc sets. I'll glad if you can try and let me know if it works for you, or what I'm doing wrong, or move the settings into a plist file.

tonklon:
appledoc 2.0.5 was indeed broken, but was already fixed [again](tomaz/appledoc/issues/107). I will prepare a settings file.

PR #727, opened 2011-11-29
Removing build warnings

yosit:
three20 includes 27 build warning out of the box - this commit removes most of them.

PR #741, opened 2012-01-06
Compiler warnings fixes

aporat:
most build warning fixes are taken from #727


diederich:
Never mind, this was fixed in the second patch...

PR #743, opened 2012-01-06
Additions to TTGlobalUICommon

sumchattering:
1) Added support for detecting iOS 5.0 and 4.3 in TTOSVersionIsAtleast() method

2) Added method IsMultitaskingSupported to check if multitasking is supported

3) Added methos TTRuntimeOSVersionIsAtleast which checks for the OS version at runtime.(useful when say iOS SDK 5.0 compiles apps are run on iOS 4)


aporat:
this is great! thank you for submitting this pull request!

(tested & merged into development)

PR #748, opened 2012-01-11
Fix scrollview offset restore

diederich:
when restoring the tableView offset, the maximum offset
can be a number < 0, if the contentSize doesn't fill the
whole tableview. If that's the case, set the maximum scrollOffset
to 0.


aporat:
nice, i think i had a similar issue with scollview offsets starting on the wrong positions. hopefully it will fix my issue as well.

btw, maybe it's better to have the feature / bug fixes branches on our own forks, so we won't have too many branches on the main repo :-)

PR #749, opened 2012-01-12
[UI] Reload buttons in TTTableViewController error views

aporat:
This pull request adds automatic reload buttons into the `TTTableViewController` error views (`TTErrorView`)

These error views are displayed when `TTTableDataSource` could not be loaded (usually happens because of internet connectivity issues). This new reload button will try and reload the `TTTableDataSource` when pushed.

Currently, if the user encounter a `TTErrorView` error message, the user either has to close & reopen the app, or rely on a refresh button (which has to be added manually by developers on a toolbar or the navigation bar)

The reload button style is also customizable using a `TTStyleSheet` 

[](http://s14.postimage.org/c33hi6pqp/Screen_Shot_2012_01_12_at_1_27_19_AM.png)

[](http://s8.postimage.org/yysff46hh/Screen_Shot_2012_01_12_at_1_34_53_AM.png)


tonklon:
Great idea, but I personally do not like the fact, that it is hard coded and cannot be switched off. It should have a switch to turn off. I'm fine with on as default.

## Rationale

I use TTTableViews together with dataSources and Models that do not come from the web, and/or do not support reloading. And although it might make sense in the vast majority of cases the just hit reload, there could be errors, for which reload just doesn't make sense.

And I am using TTErrorView in different places then TTTableViewController. That might be my fault, because the directory structure say it belongs to the table View.

This is only my personal opinion, don't know how others think about it.

aporat:
good points. thanks.

I'm thinking it might be good to have a generic `TTErrorView` (and move it away from the Table Views folder) and add a `TTTableErrorView` which extends  `TTErrorView`.

Also, you're right. it should have a way to disable the refresh button. an optional function on the data source (when we set the error messages and images) would work perfectly, but I haven't got to that step yet :-)

aporat:
a function named `- (BOOL)reloadButtonForEmpty` was added in the tableview's `TTTableViewDataSource` which allows developers to disable the reload button if they don't wish to have it in the error views

PR #751, opened 2012-01-16
add Xcode 4 workspace

diederich:
This pull requests add an xcode4 workspace to the repository. Currently it just holds all of three20's projects and provides 2 schemes. One to build all the demos and another to run all of the unit tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,6 @@ Products
 *.pbxuser
 *.perspectivev3
 *.mode1v3
-*.xcworkspace
 xcuserdata
 .DS_Store
 config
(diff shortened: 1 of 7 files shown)


aporat:
wow, this is really cool feature. I'll definitely use this with three20 and other projects.

I tried using this workspace locally, and there's alot of "junk files" (mostly scripts & docs) inside the workspace. maybe we should leave these files outside the workspace and have only the actual three20 projects?

diederich:
Glad you like it. It's really quite useful for lager projects. 
Note though, this is not really how it's meant to be used. Projects don't need to be sub-projects of each other - projects within the same workspace can be linked together easily. It seems to me that support is improved in every Xcode release, so we maybe just need to be patient ;)
Anyways, this is something to begin with.

Regarding the "junk files": I've added them to have everything in one place. If this adds too much clutter, we can hide them in folders/groups...

aporat:
merged into 1.0.11-dev

PR #752, opened 2012-01-17
Using TTMask on iOS 5.0.1 results in upside down images being drawn. Rem...

fleitz:
...ove code to invert the context.

### DIFF
--- a/src/Three20Style/Sources/TTMaskStyle.m
+++ b/src/Three20Style/Sources/TTMaskStyle.m
@@ -66,8 +66,9 @@
     CGContextSaveGState(ctx);
 
     // Translate context upside-down to invert the clip-to-mask, which turns the mask upside down
-    CGContextTranslateCTM(ctx, 0, context.frame.size.height);
-    CGContextScaleCTM(ctx, 1.0, -1.0);
+    // Turning the mask upside down is no longer necessary. 
+    //CGContextTranslateCTM(ctx, 0, context.frame.size.height);
+    //CGContextScaleCTM(ctx, 1.0, -1.0);
 
     CGRect maskRect = CGRectMake(0, 0, _mask.size.width, _mask.size.height);
     CGContextClipToMask(ctx, maskRect, _mask.CGImage);


diederich:
Thanks!

IIRC, this changed between iOS versions. Do you know when this changed, and shouldn't we make this work on older versions, too? If the old version is not needed anymore, you can remove the code instead of commenting it out.

Also, please make sure to target the development branch with pull requests.

fleitz:
No idea when it changed. Today was the first time I've used
ttmaskstyle. Unfortunately I don't have any older iOS devices to test
on.

Sent from my iPad

On Jan 17, 2012, at 2:49 AM, Stephan Diederich
reply@reply.github.com
wrote:

aporat:
sounds like a really odd bug. 

It would be great if you can post your code here, so we can try and replicate this issue

fleitz:
Here's the style I'm using:
- (TTStyle_) coloredIcon:(UIColor_) color{
  
  return
  
  [TTMaskStyle styleWithMask:[UIImage imageNamed:@"Housing-Icon"] next:
  
    [TTContentStyle styleWithNext:
  
     [TTBlendStyle styleWithBlend:kCGBlendModeMultiply next:
  
  ```
  [TTSolidFillStyle styleWithColor:color next: nil]]]];
  ```

}
- (TTStyle*) amberIcon {
  
  return [self coloredIcon:RGBACOLOR(255, 255, 0, 1.0)];

}

And here's the context in which the style is used:

[TTTableImageItem itemWithText:self.opportunity.name imageURL:
@"bundle://Housing-Icon.png" defaultImage:nil imageStyle:TTSTYLE(amberIcon)
URL:nil]];

The idea is to mask the kCGBlendModeMultiply so I can color only the
non-transparent parts of the image. I think there is actually a larger
issue with masking that's not apparent because the image in question is
circular.

Here's a link to the "Housing Icon" image: http://i.imgur.com/bv2Es.png

On Tue, Jan 17, 2012 at 12:40 PM, Adar Porat <
reply@reply.github.com

diederich:
Sorry for the misleading comment above. The flip of the context is necessary, as the coordinate system of Quartz is flipped and thus the mask needs to be flipped, too. You can find a working example of this in TTCatalog's styledViews part (StyleTestController - a head colored with a gradient).

While I'm not really experienced in Quartz drawing, I just played with this a bit and here's an example that seems to do what you want:

``` objective-c
 [TTImageStyle styleWithImage:TTIMAGE(@"bundle://bv2Es.png") 
                 defaultImage:nil 
                  contentMode:UIViewContentModeBottomLeft 
                         size:CGSizeMake(100,100) next:
    [TTMaskStyle styleWithMask:TTIMAGE(@"bundle://bv2Es.png") next:
     [TTBlendStyle styleWithBlend:kCGBlendModeMultiply next:
      [TTSolidFillStyle styleWithColor:RGBACOLOR(255, 255, 0, 1.0) next:nil]]]],
```

The idea is to draw the image, set the blend mode and then color the non-alpha parts of the image with a solid fill style. (workflow stolen from http://coffeeshopped.com/2010/09/iphone-how-to-dynamically-color-a-uiimage after a quick web search).

[closing this pull requests, as images as well as masks need to be flipped when drawn with CG]

PR #746, opened 2012-01-11
iOS 5 ipad compatibility issues

aporat:
this pull request includes some changes to TTBaseController which solves the issues with UIPopViewController & TTSplitViewController under iOS 5.

The parentViewController function role was under with iOS 5 (http://omegadelta.net/2011/11/04/oh-my-god-they-killed-parentviewcontroller/) which caused issues with TTSplitViewController when the secondary view controller is displayed as a popup view controller.

closes #713

PR #754, opened 2012-01-20
Safeguard against UITabBarController selecting  a controller not in the tab bar

kemenaran:
Apple Docs states that only controllers in -[UITabBarController viewControllers] can be set as the -selectedViewController.

If for some reasons iOS messes up the view controllers hierarchy (it happens), setting an unknown view controller will fail (and present a blank screen). This fix avoids that.

### DIFF
--- a/src/Three20UI/Sources/UITabBarControllerAdditions.m
+++ b/src/Three20UI/Sources/UITabBarControllerAdditions.m
@@ -89,7 +89,9 @@ TT_FIX_CATEGORY_BUG(UITabBarControllerAdditions)
 
 ///////////////////////////////////////////////////////////////////////////////////////////////////
 - (void)bringControllerToFront:(UIViewController*)controller animated:(BOOL)animated {
-  self.selectedViewController = controller;
+  if ([self.viewControllers containsObject:controller]) {
+    self.selectedViewController = controller;
+  }
 }
 
 


tonklon:
I don't like that fix. IMHO in a situation like this an Exception should be thrown (internal Inconsistency). But if the OS interferes here this is not an option. Do you have an example to reproduce the problem?

kemenaran:
I wrote an example when opening the Pull Request, and decided to keep it aside fro brevity's sake :) Anyway,I kept it somewhere, so here it is :

(see below, wrong button clicked.)

kemenaran:
An example of messed up parentViewController hierarchy can be the following hierarchy of view controller :

```
UITabBarController
  -> UIViewController1
    -> UINavigationController(Modal)
      -> UIViewController2
```

In this case, on iOS < 5, the parentViewController of UINavigationController is the UITabBarController — instead of the expected UIViewController1. iOS >= 5 has the expected behavior though.

This causes Three20 to try to set UINavigationController as the selected controller of UITabBarController (when the TTNavigator loops over newly created controller to ensure they are actually visible, by sending -bringControllerToFront: messages recursively).

aporat:
@kemenaran do you mind trying pulling #746 into your fork and test if that patch solves your issue (without your change)? 
 sounds like another parentViewController issue to me

jamesgpearce:

PR #756, opened 2012-01-21
Fix TTLayoutManager analyzer warnings

diederich:
see #753
- imageSize stuff is rewritten to make the analyzer happy
- frameWidth is always set to something before it's used, so removing that line should be safe


diederich:
oops, wrong branch. sorry for the noise...

PR #758, opened 2012-01-21
TTTableItem / TTTableItemCell table cell mapping reorganization  

aporat:
Based on pull request #293. 

Instead of having `TTTableItem` <> `TTTableItemCell` mapping hardcoded inside the `TTTableViewDataSource` class, each `TTTableItem` subclass now contains his own cell class mapping.

This change is backward compatible with apps that use custom TTableItemCell subclasses (original pull request broke any app with custom TTableItemCell)


diederich:
Thanks for bringing this up again. 

Did you have a look at my branch https://github.com/diederich/three20/tree/three20/remove-hard-coded-cell-mapping ?
It hold the original author's commits and on top of that a few changes.

aporat:
that's cool. I like the fact you used a protocol. it looks like a better way to implement these changes.

My biggest issue with the original pull request is that it broke my apps that used custom TTTableItem subclasses.

 I couldn't test your branch, because it's a little behind, but I think it will throw exceptions instead of silently break.

diederich:
Maybe we can combine the two approaches... I've rebased my feature branch at https://github.com/diederich/three20/tree/three20/remove-hard-coded-cell-mapping on top of current development

aporat:
sure, I'll try to look at it in the next few days, but your branch still breaks my existing apps that subclass TTTableItem.

PR #762, opened 2012-01-24
TTTableItem / TTTableItemCell table cell mapping reorganization   

aporat:
Based on pull request #293. 

Instead of having `TTTableItem` <> `TTTableItemCell` mapping hardcoded inside the `TTTableViewDataSource` class, each `TTTableItem` subclass now contains his own cell class mapping.

This change is backward compatible with apps that use custom TTableItemCell subclasses (original pull request broke any app with custom TTableItemCell)


diederich:
ouch. something bad happened in that pull request, duplicate commits, unrelated changes. closing...

diederich:
see https://github.com/diederich/three20/tree/remove-hard-coded-cell-mapping for an updated version.

(and for initial discussions: #293)

PR #763, opened 2012-01-24
line 93~95 will fail when "destW" or "destH" is not integral.

kimdwkimdw:
Like this "CGBitmapContextCreate: unsupported parameter combination: 8 integer bits/component; 32 bits/pixel; 3-component color space; kCGImageAlphaNoneSkipLast; 3199 bytes/row."!

CGSize *size = image.size; //width : 600px
size.width can make 599.998


jamesgpearce:

PR #767, opened 2012-02-02
improvement for fieldName

XBeg9:
I have need in custom fieldName and filename option in addFile function. This fix has improved generatePostBody

### DIFF
--- a/src/Three20Network/Sources/TTURLRequest.m
+++ b/src/Three20Network/Sources/TTURLRequest.m
@@ -240,12 +240,20 @@ const NSTimeInterval TTURLRequestUseDefaultTimeout = -1.0;
     NSData* data = [_files objectAtIndex:i];
     NSString* mimeType = [_files objectAtIndex:i+1];
     NSString* fileName = [_files objectAtIndex:i+2];
-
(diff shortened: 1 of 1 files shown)


diederich:
hm, looks like this breaks the internal logic of the _files array, if there is more than one file.

Currently the _files array seems to expect exactly 3 parts: (data, mimeType,fieldName). See the loop here, especially "i+3" https://github.com/XBeg9/three20/blob/670d765990ad189893d139866f2c7efb5c3dc176/src/Three20Network/Sources/TTURLRequest.m#L239

I guess this could be fixed by either adding always 4-parts (and thus an empty element if there's no fieldName), or e.g. by bundling the information in the array either in a custom object, or by using a dictionary.

what do you think?

XBeg9:
Yes, I just show you my "fix" :) I am new in iOS development

diederich:
Yeah, thanks for taking the time to do this! 
Feel free to update this pull request with an adjusted fix!
(+1 for creating a new pull request, targeting the development branch :))

XBeg9:
ok, thanks:) have a nice day :)

PR #742, opened 2012-01-06
Project Modernization with Xcode 4.2

sumchattering:
Should we do this.? I see you guys havent done this..Why not?

PR #768, opened 2012-02-03
LLVM ARM6 release build optimizer bug...

troppoli:
We have bumped into an issue that apparently a lot of other people have.... some three20 users, some not. Basically LLVM+ARM6 and thumb optimizations don't mix. It would seem that the best thing to do is to turn off thumb support for the time being to avoid the trouble. In our case, text was all over the place due to 
TTTextStyle's (CGRect)rectForText:(NSString_)text forSize:(CGSize)size withFont:(UIFont_)font returning bogus values.

See....

http://www.idevgames.com/forums/thread-9606.html

http://stackoverflow.com/questions/9009883/weird-three20-graphics-bug-in-thumb-viewer-when-running-in-release-mode


aporat:
I just found out this nasty bug happens on my a few of my published apps as well, as I heavily use TTButton + TTStyleSheets. I guess this bug was introduced since the recent move to LLVM in #742.

diederich:
Thanks of looking into this and finding the right place to put it in ;)
I have this enabled in my apps which still need to support armv6, too.

troppoli:
I pulled to the head of dev and verified it yesterday before applying the fix.

PR #775, opened 2012-02-07
Apply default cell selection style in TTTableViewCell's initializer 

diederich:
(I have this commit lying around side quite a while, and since Three20Style is already a dependency of Three20UI I see no reason this can't be applied)


jamesgpearce:

PR #776, opened 2012-02-07
adjust header locations

diederich:
[Build] adjust and simplify header handling
public headers from Three20's static libraries now
go to $(TARGET_BUILD_DIR)/include/$(PRODUCT_NAME).
We choose this folder, as Xcode automatically adds "$(TARGET_BUILD_DIR)/include"
to HEADER_SEARCH_PATH of a target.
Thus includes in your apps of the form "#import "Three20UI/Three20UI.h" just work™.

Unfortunately this does not happen for Archive builds, so we need to do it manually in that case:
"$(OBJROOT)/UninstalledProducts/include" (see src/common/Configuration/Paths.xcconfig)

The old way of moving the headers out of Release-xxx/Debug-xxx folders is bad:
first it doesn't allow Xcode to clean the headers during a "clean project" and second,
we might have different(ly processed) headers for Debug and Release builds.

This is probably a breaking change for existing apps and their "Archive" phase. Still I'm sending this pull request to think about it and to get some feedback....


jamesgpearce:

PR #777, opened 2012-02-07
don't swap Tableview's datasource

diederich:
This patch removes swapping away the datasource of a tableview while the model is not shown. Instead it just sets the tableView's hidden property to YES.

Removing and re-adding the datasource is pretty much just a bag of hurt. E.g. it can happen that during model changes, the datasource gets replaced. If this happens, the tableView's beginUpdates and endUpdates work with different datasources (and thus crashes pretty much all the time).


jamesgpearce:

PR #784, opened 2012-03-08
Fix popover crash in navigator

kemenaran:
When the TTBaseNavigator creates a new popover, it does it using the `-init` message.

This throws an exception on iOS 4.3 and iOS 5.1: the UIPopoverController designated initializer is actually `-initWithContentViewController:`.

The fix uses the correct message, and thus prevents the crash.


tonklon:
Thanks a lot.

PR #789, opened 2012-03-22
Add support for XML parsing through GDataXML

mkhurramali:
I have added support to extThree20XML for parsing XML documents using GDataXML.


diederich:
Nice, thanks. Any comment on why to use GDataXML?

While I haven't done much with the TTExtensions yet, it looks to me that in the other extension (e.g. JSON) different implementation options are handled differently to your pull request. 
extThree20JSON has different targets to link against. E.g. there's extThree20JSON+SBJON.a if you want the SBJSON implementation, and extThree20JSON+YAJL.a for YAJL. While it might seem a bit complicated, it allows users to choose one or the other without changing Three20 source itself.

mkhurramali:
The reason I like GDataXML is that it allows me to have a very similar code structure when reading a JSON document using extThree20JSOn or XML document using my modifications.

I have now fixed the structure to closely follow extThree20JSON.

diederich:
Thanks for adopting the structure of the JSON project. 
Besides the comments in the commit, this looks all good to me...

mkhurramali:
Thanks Stephan, for your time and valuable feedback. I have made fixes according to all your comments, except this one:

Seeing this, does it make sense to rename the other target, too?

I could not find what line of code this comment was based on.

If everything is good, do I need to do anything else for this pull request to complete?

diederich:
Thanks for applying the changes and sorry for the late reply. 
"Seeing this, does it make sense to rename the other target, too?" was just me being confused about the different xcconfigs. Once I figured it out, I deleted the comment ;)
The pull request looks good to me.

mkhurramali:
Closing this request as I have sent the same request to the development branch.

PR #806, opened 2012-07-19
Invalidate the `_editHoldTimer` before releasing `_buttons`

mattrubin:
Invalidate the `_editHoldTimer` before releasing `_buttons` to avoid the crash which occurs when `editHoldTimer:` is called after the button which triggered it has been deallocated. If `setPages:` or `setColumnCount:` are called between when the button is touched and when the timer goes off one second later, the button the timer references will have been deallocated, leading to a crash.


jamesgpearce:

PR #817, opened 2012-09-21
Fixes for iOS 6 and building with Xcode 4.5

ettore:
- Updated architectures to be armv7 and armv7s (dropped armv6). Currently allowing armv7 as the only Valid Architecture, since iPhone 5 has not been tested yet. (Once it is, all one needs to do is to flip Valid Architectures to "armv7 armv7s".) 
- Removed Target level flag for PIC (position independent code) since it's already set at the Project level.
- replaced deprecated enums and fixed keyboard rotation geometry calculations
- commented out hacky references to UITouch ivars, now removed by Apple. These were mostly used in debugging.
- removed old ass __IPHONE_3_21 ifdefs, defined in all iOS SDKs since several years.
- fixed some other Xcode warnings


diederich:
Thanks for the PR. Three20 uses xcconfigs for easier build system configuration (especially diffs), so changing the values there instead of the xcconfigs would be appreciated.

Regarding architectures, I'm not sure to drop armv6. Three20 is probably used in lots of legacy projects, so as long it's not needed (or still allowed to build for it), maybe leave it there?
I'm not sure about the VALID_ARCHS vs. ARCHS changes. Isn't VALID_ARCHS for all the archs that are buildable, while ARCHS specifies the ones that should actually be built for the current configuration/SDK?

I'd say we can remove the ivars instead of commenting them. Should be OK to leave them only in the Git history :)

I've just merged another PR that only had the project updates. You might want to rebase on current development, see what's left resubmit (force push) this PR again.

ettore:
xcconfigs: got it, I wasn't aware of them.

armv6: ok, I just disabled it in my branch. I'm still not sure how any app built with Xcode 4.5 (the only one that can target iOS 6 and therefore submit to Apple) can make use of a armv6 binary. Even supposing we're building three20 as a library and linking it to our app, Xcode 4.5 will complain about the armv6 binary since it can't build the app in that arch. Most likely this is the situation most users will run in, so in my opinion Three20 should cater to that? But I guess it's not my decision to make. 

good point about VALID_ARCHS vs ARCHS. 

Yeah, not sure what's left here then... but supposing I were to resubmit this same PR (and not open a new one), how would I do that? I guess I'll need to move my branch origin to later in the future? I'm kind of unfamiliar in this process.

diederich:
Hm, not sure there's an issue with having armv6 in the library binary. The linker, when assembling the final app binary, should one pick those architecture out of the .a file that it needs in the executable. So if the app is built armv7 only, it should only take the armv7 part out of three20. Anyways, I guess armv6 must go in the near future.

---

It looks like you have a 'feature' branch in your repository named ios6_fixes. IIRC you should be able to update this pull request by just updating your branch in your repository. As this is going to change 'history' of the branch, you'll need to force push it. 
Should look something like this:

```
git fetch --all (brings in new three20 development changes)
git checkout ios6_fixes 
git rebase -p origin/development (rebase the ios6_fixes branch on top of current development. This should be the most work. Some commits won't apply, some are already there, so you need to tell git to skip them,....)
```

Now there should be a shiny new ios6_fixes branch that (while having nothing to do with the last branch) can be (force) pushed to your repo. If all goes well, this PR should be updated :)

ettore:
Thanks so much for the instructions. I might just abandon this pull request since I already merged the feature branch on my main branch, and AFAIK rebasing is bad in this situation.

PR #818, opened 2012-09-22
Fix all remaining compiler warnings

cliffrowley:
Since 4.3 was removed from Xcode under Mountain Lion I pretty much use the latest iOS SDK, so I don't fully appreciate the effect a couple of the changes might have for earlier SDKs (NSXMLParserDelegate).

Feel free to advise / reject / ignore.

Sure is nice seeing a build with 0 errors, 0 warnings, though.


diederich:
NSXMLParserDelegate should have no effects, besides compiler errors when building against something < 4.0.

Merged. Thanks!

PR #819, opened 2012-09-25
fixed a few iOS 6 deprecation warnings

ettore:
Removed call to setStatusBarHidden:animated:, deprecated in iOS 6.
Removed some old code wrapped in __IPHONE_3_2 macro, since Three20 now targets iOS 4+.
Added pragma to ignore deprecated declarations in code handling keyboard notifications.


ettore:
Actually, setStatusBarHidden:animated: was deprecated in iOS 3.2. My bad.

diederich:
Thanks! I'm not really sure about pragma'ing away those Keyboard notification warnings, but.. yeah :)

PR #821, opened 2012-09-26
fixed background color not being set for grouped tableviews on iOS 6

ettore:
Grouped-styled tableviews appear with the default striped background even when a background color is set. This happens only on iOS 6. 

This fix solves it by setting to nil the backgroundView property. Another way to do this would have been to set the opaque property to NO, which seemed more inefficient.

While at it, I removed repeated ifs and made the code a little easier to read.


jamesgpearce:

PR #824, opened 2012-10-16
Fix build errors on real devices.

infocube:


diederich:
Hi infocube, thanks for the PR.

Did you check if there's already a setting in one the xcconfigs that could be used to build armv7 and armv7s?

infocube:
Hi,

Just fixed it, config in one file for all projects.

diederich:
Hm, so you're dropping armv6 in favor of armv7s support?

Is armv7s already an issue? Three20 is probably used in 'legacy' projects, so a missing armv6 might be an issue for more users. What do you think?

daveverwer:
I am indifferent on the armv7 support but dropping armv6 is essential and the project currently does not compile for devices using Xcode 4.5 as armv6 compilation is no longer supported.

PR #828, opened 2012-12-20
UITouch extensions don't compile by default in Debug configuration

cglong:
The Three20 `UITouch` extensions break when using the iOS 6 SDK with the `DEBUG` flag set. This patch will cause the extensions to not be compiled by default in the Debug configuration; they must now be specified using the `DEBUG_TOUCHES` flag.

This patch comes wholly from @kemenaran in cogenta/three20-1.0.5@be70e1ff3965c0ff4f7598bf5213349629a27540.


jamesgpearce:

PR #832, opened 2013-05-28
remove all sources and update README

diederich:
This PR removes all sources and updates the README to state that three20 is not maintained anymore.

It seems there are still new iOS developers that pick up three20 from somewhat old blog posts or tutorials and then ask questions about it on stackoverflow. By removing all sources and updating the readme it should become clear that three20 is at the end of its lifetime.

I'd propose to merge this PR into development-branch, after that merge the new development branch into master. 
Optionally, one could do a final release (merge development to master, tag 1.0.12). If someone feels like maintaining or merging one of the open PR, this commit could ofc be reverted on development later.

What do you think?


aporat:
:+1: for the heads up for new developers

I suggest deleting the development branch completely, and add a huge note on the master's README file saying "three20 is no longer maintained" (but keep the entire README content as is). if you have access to the repo's admin, it also would be nice to remove the issues tab, so developers won't be able to open issues.

a warning on the README should be enough in my opinion. i don't see a reason to delete the entire source code from the master branch. it doesn't hurt anyone as long as the warning is big enough.

diederich:
Thanks for the feedback.

I'm not sure a warning in the README is enough, people seem to miss warnings from time to time, and ofc it can be easily missed by updating only a submodule. Having no source anymore should be quite clear :-)
Is there a reason to keep the README and the source?

It might be a good idea to make the issues read-only, but I'm not sure this is possible. It also doesn't hurt that much, at least if no one is spammed by the mails :-)

Regarding development branch, I guess you're right. If people still need to fix stuff, they can do it in their own clones.

http://github.com/Three20 is also still open and should probably be closed down. @jverkoey are you in control of it?

phillipuniverse:
+1. As much as it makes me sad to see it go this seems like the most appropriate option so people don't continue to use non-maintained code.

I am for deleting the source code on the main branch in GH (master, currently) so that it's very obvious that it's no longer maintained. Here is what I would do:
- Merge in #833 into master
- Update the README on all branches saying that it's no longer maintained (keep code and remainder of README intact)
- Make a new branch off of 'master' called something like 'non-maintained' I dunno
- In the README of the 'non-maintained' branch, delete all source code and in the README tell people that this project is no longer maintained by anyone
- Change the settings in the GH repo such that the 'non-maintained' branch is the new default. This ensures that people who go to github.com/facebook/three20 see the deleted source code first.

Someone should also update http://three20.info to make this really obvious as well. I assume @jverkoey owns that?

One other nice thing would be to let people know alternatives to Three20 (Nimbus and AFNetworking come to mind) that have been kept up.

diederich:
@phillipuniverse Thanks for the input.

What we basically want to achieve is that _everyone_ gets three20 isn't maintained anymore. This should be clear when 
1) going to the facebook/three20 github page
2) someone updates their git submodules.

1) can be achieved by cleaning up a branch and setting it as the default branch. (either master, or the proposed non-maintained)
2) must be done on the branch most people have checked out. So this is probably master or development. 

I've updated the README.md to point to Nimbus (and to the migration guide), as well as how to get back to the version before we removed all sources.

Recap:
- merge #833 to master to mark the final 'release'
- merge this PR to shut-down development-branch
- merge development to master (to cleanup master as well)
- (optional: if someone really wants to do further bugfixing, revert this PR merge on development after merge to master, so that development can just go on :-))

diederich:
:metal:

rtwomey:
happy_dog

